PR #4, opened 2015-07-16
Do not shutdown a glidein if it is claimed/idle.

bbockelm:
If a glidein is in the claimed/idle state, we shouldn't shut it down.  It could just be an indicator that the schedd has a long queue for launching shadows - in such a case, shutting down the glidein will only increase churn and the queue length.


mambelli:
Accepted and included in 3.2.11
See https://cdcvs.fnal.gov/redmine/issues/9646

PR #8, opened 2016-08-15
Update setup_x509.sh

efajardo:
The correct code to compare if binaries are in the path is 127 and not 1.

### DIFF
--- a/creation/web_base/setup_x509.sh
+++ b/creation/web_base/setup_x509.sh
@@ -87,7 +87,7 @@ function check_x509_tools {
     missing_commands=0
     # verify grid-proxy-info exists
     command -v grid-proxy-info >& /dev/null
-    if [ $? -eq 1 ]; then
+    if [ $? -eq 127 ]; then
 	STR="grid-proxy-init command not found in path!"
 	echo $STR >&2
         #"$error_gen" -error "setup_x509.sh" "WN_Resource" "$STR" "command" "grid-proxy-init"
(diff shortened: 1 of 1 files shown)


holzman:
Hi Edgar,

What version of bash are you using? My bash manpage says:
> If the -V or -v option is supplied, the exit status is 0 if command was found, and  1  if  not.

And I verified that from the CLI.

efajardo:
Hi @holzman ,

I tried this in centos and suse and other redhat and it should 127:

```
[0726] efajardo@uaf-7 ~$ grid-proxy-init-2
-bash: grid-proxy-init-2: command not found
[0726] efajardo@uaf-7 ~$ echo $?
127
```

Moreover doing a quick google search:

Doing a quick google search agrees with me:

http://tldp.org/LDP/abs/html/exitcodes.html

Moreover I have found that the script returns command not found when actually grid-proxy-info just failed:

```
[0731] root@osg-ligo-1 ~#  grid-proxy-info

ERROR: Couldn't find a valid proxy.
       globus_sysconfig: Could not find a valid proxy certificate file location
globus_sysconfig: Error with key filename
globus_sysconfig: File does not exist: /tmp/x509up_u0 is not a valid file

Use -debug for further information.
[0731] root@osg-ligo-1 ~# echo $?
1
```

holzman:
Hi Edgar,

I'm still not getting it. The script executes `command -v grid-proxy-info >& /dev/null`, not grid-proxy-info directly, right?

efajardo:
Hmm yea I think I was doing somehting wrong, you can ignore this pull request.

PR #9, opened 2017-01-18
Update cgWCreate.py

efajardo:
The projectName for HTCOndor-CE was missing the double quotes.


efajardo:
@mambelli could you review this and merge please?

PR #10, opened 2017-02-03
Fix bug introduced in 3.2.16 when several authentication methods are used

efajardo:
Looking at the actual printout:

`[2017-02-03 13:27:35,436] DEBUG: glideinFrontendInterface:448: am_set in support_auth_method: set(['project_id', 'grid_proxy'])
[2017-02-03 13:27:35,436] DEBUG: glideinFrontendInterface:449: type_set in support_auth_method: set([u'grid_proxy'])`

It is was clear the comparison was not correctly done.


efajardo:
It was actually well performed the comparison before.

PR #12, opened 2017-11-07
Update URL for epel-release RPM

matyasselmeci:
The old location for the epel-release RPM no longer works; changed it to a new URL that is a symlink that is guaranteed to be updated.


mambelli:
Accepted (1d3236a2f) and included in version 3.2.21.

PR #13, opened 2017-11-30
Provide script to handle frontend proxy renewal

brianhlin:
https://jira.opensciencegrid.org/browse/SOFTWARE-3022

Provide a built-in service for frontend admins to renew their frontend and pilot proxies. Prefer `voms-proxy-fake` to prepare for VOMS Admin server retirement in the OSG.

CC: @bbockelm @efajardo for comments


build/packaging/rpm/glideinwms.spec
```diff
@@ -251,6 +252,7 @@ install -m 0500 frontend/manageFrontendDowntimes.py $RPM_BUILD_ROOT%{_sbindir}/
 install -m 0500 frontend/stopFrontend.py $RPM_BUILD_ROOT%{_sbindir}/stopFrontend
 install -m 0500 frontend/glideinFrontend.py $RPM_BUILD_ROOT%{_sbindir}/glideinFrontend
 install -m 0500 creation/reconfig_frontend $RPM_BUILD_ROOT%{_sbindir}/reconfig_frontend
+install -m 0500 frontend/renew_proxies.py $RPM_BUILD_ROOT%{_sbindir}/renew_proxies
```
bbockelm: Do we realistically expect a sysadmin to run this script by hand?  If so, this is the right directory.  Otherwise, something like `libexec/%{name}` might be better.  

I would prefer some prefix to make it obvious that this command is associated with gwms ... but it seems it's not done elsewhere?  (aside: holy buckets, there's no discernible naming convention here!)

build/packaging/rpm/glideinwms.spec
```diff
@@ -297,11 +299,17 @@ rm -f $RPM_BUILD_ROOT%{python_sitelib}/glideinwms/creation/templates/factory_ini
 rm -f $RPM_BUILD_ROOT%{python_sitelib}/glideinwms/creation/templates/frontend_initd_startup_template_sl7
 rm -f $RPM_BUILD_ROOT%{python_sitelib}/glideinwms/creation/templates/gwms-factory.service
 rm -f $RPM_BUILD_ROOT%{python_sitelib}/glideinwms/creation/templates/gwms-frontend.service
+rm -f $RPM_BUILD_ROOT%{python_sitelib}/glideinwms/creation/templates/gwms-renew-proxies.cron
```
bbockelm: I'm assuming we have to remove this due to a poorly written `setup.py`?

(this RPM .spec file is really quite the monster, but cleaning it is out-of-scope for this PR...)

config/proxies.ini
```diff
@@ -0,0 +1,61 @@
...
+proxy_key = /etc/grid-security/gwms-frontend/pilotkey.pem
+
+##################################
+# Virtual Organization information
+# vo:
+#     name of the virtual organization
+# fqan (optional, default: /Role=NULL/Capability=NULL):
+#     Request FQAN VO attribute
+# use_voms_admin (optional, default: false):
+#     If true, contact the VO's VOMS Admin server to sign the VOMS attributes
```
bbockelm: This attribute is misnamed -- technically, it's the "VOMS server" (which, in 2.x, is written in C) while VOMS Admin (written in Java) is the web frontend.

Might as well get it correct to start out with...

creation/templates/gwms-renew-proxies.timer
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Hourly frontend and pilot proxy renewal
+After=syslog.target network.target
+Wants=network.target
+
+[Timer]
+OnActiveSec=0s
+OnUnitActiveSec=1h
+Persistent=true
```
bbockelm: I typically request that `RandomizedDelaySec` to be set for timer units (I'd suggest 5 minutes is appropriate here).  That helps prevent things from synchronizing across multiple clients.

Here, I expect there are so few clients it seems pretty optional.  Couldn't hurt?

bbockelm: Also, from the docs for `Persistent`:

> Note that this setting only has an effect on timers configured with OnCalendar=.

So please remove this line.

frontend/renew_proxies.py
```diff
@@ -0,0 +1,151 @@
...
+    if len([x for x in proxies if x.startswith('PILOT')]) < 1:
+        raise RuntimeError("ERROR: there must be at least one [PILOT] section in %s" % CONFIG)
+
+    # Proxies need to be owned by the 'frontend' user
+    try:
+        fe_user = pwd.getpwnam('frontend')
+    except KeyError:
+        raise RuntimeError("ERROR: missing 'frontend' user")
+
+    # Load VOMS Admin server info for case-sensitive VO name and for faking the VOMS Admin server URI
```
bbockelm: Hm - at some point, we'll need to think about how we get rid of the URI... I suppose that's not for today.

frontend/renew_proxies.py
```diff
@@ -0,0 +1,151 @@
...
+                vo_attr.cert = proxy_config['vo_cert']
+                vo_attr.key = proxy_config['vo_key']
+                stdout, stderr, retcode = voms_proxy_fake(proxy, vo_attr, voms_info['uri'])
+        else:
+            print "WARNING: Unrecognized configuration section %s found in %s.\n" % (proxy, CONFIG) + \
+                "Valid configuration sections: 'FRONTEND' or 'PILOT'."
+
+        if retcode == 0:
+            proxy.write()
+            os.chown(proxy.output, fe_user.pw_uid, fe_user.pw_gid)
```
bbockelm: This has a race condition that may end in tears.  Do this instead:

1.  Create the proxy in a temporary filename (preferably using a class from the `tempfile` library).  Should be in the same directory as the output in order to avoid cross-filesystem renames later (keeps the rename atomic).
2.  Do the appropriate `chown`
3.  Write out the proxy, fsync the file descriptor, close the fd.
4.  Use `os.rename` to overwrite the new proxy with the old one.

This way, you're guaranteed that the file at `proxy.output` always is valid.

frontend/renew_proxies.py
```diff
@@ -0,0 +1,151 @@
...
+            print "Renewed proxy from '%s' to '%s'." % (proxy.cert, proxy.output)
+        else:
+            # don't raise an exception here to continue renewing other proxies
+            print "ERROR: Failed to renew proxy %s:\n%s%s" % (proxy.output, stdout, stderr)
```
bbockelm: The program should return non-zero if any proxy renewal failed.

frontend/renew_proxies.py
```diff
@@ -0,0 +1,151 @@
...
+        except ValueError:
+            return 0
+
+class VO(object):
+    """Class for holding information related to VOMS attributes
+    """
+    def __init__(self, vo, command, cert=None, key=None):
+        self.name = vo
+        command = command.lstrip('/')
+        self.fqan = '/%s/%s' % (vo, command)
```
bbockelm: This looks wrong.  Wouldn't it end up with `/osg/osg/Role=pilot` when the intent is `/osg/Role=pilot`?

However, as a sanity check, I think you *should* ensure that for `VONAME`, if a FQAN is present, it should start with the prefix `/VONAME`.

I.e., an attempt to create a proxy named `/foo/Role=pilot` for the VO `osg` should end in failure.

build/packaging/rpm/glideinwms.spec
brianhlin: It's even worse than that: https://github.com/holzman/glideinWMS/blob/master/build/packaging/rpm/glideinwms.spec#L268

frontend/renew_proxies.py
brianhlin: Isn't 3 already handled by 1? I can give the various `voms-proxy-*` commands the tempfile as output but I don't get choose when the fd writing/handling occurs.

frontend/gwms_renew_proxies.py
```diff
@@ -42,12 +42,13 @@ def timeleft(self):
...
+    def __init__(self, vo, fqan, cert=None, key=None):
+        if not fqan.startswith('/%s/' % vo):
+            raise ValueError('FQAN (%s) does not begin with specified VO (%s). Verify %s.' % (fqan, vo, CONFIG))
         self.name = vo
-        command = command.lstrip('/')
-        self.fqan = '/%s/%s' % (vo, command)
+        self.fqan = fqan
         # intended argument for -voms option "vo:command" format, see voms-proxy-init man page
-        self.voms = '%s:/%s' % (vo, command)
+        self.voms = fqan.replace('/%s' % vo, vo + ':')
```
bbockelm: Is this really what we want here?  Does it replace all instances?  Or do we just want the first one?

brianhlin:
@bbockelm fixed the fqan replacement issue as well as atomic writes. Please re-review

mambelli:
These scripts are completely separate from the GlideinWMS software. 
GlideinWMS includes a mechanism in the configuration to recreate a credential when there are less than N seconds to its expiration. To our fault, this was not documented. 
This pull request has the advantage to run via cron and be independent from the Frontend cycle,
The disadvantage that frontend administrators would have to write a new config file and be consistent with frontend.xml (for the proxy file location). 
The minimal change needed is to allow the frontend user to be something different from 'frontend'. You can get the correct user by checking the startup script or the owner of the main configuration script (/etc/gwms-frontend/frontend.xml in RPM installs, in the startup script for tarball installs). The rest is OK with the caveat of the 2 config files.
I'm sending to the Frontend admins an email to get some feedback before approving the pull request.

brianhlin:
Marco,

1. Did you get any feedback from frontend admins? 
1. Does the built-in gwms credential renewal handle VOMS attributes? If so, what is the plan to accommodate VOMS attribute generation with the OSG plan to retire VOMS admin servers? 
1. Is there a canonical location for the startup script that I can query?

Thanks,
Brian

mambelli:
Brian,
1. I did not get any feedback
2. The GWMS mechanism is allowing the admin to specify a command line to execute to renew the proxy but is not providing a script to do that, was expecting people to use voms-proxy-init with appropriate parameters. Your script is better because it can invoke also voms-proxy-fake and finds some of the necessary options.
3 Do you need the startup script or the config file? My suggestion is to accept an installation dir parameter (defaulting to /var/lib/gwms-frontend/vofrontend) and check the config file VOFRONTEND_INSTALL_DIR/frontend.xml for content and ownership.     Here the long explanation: the canonical location in RPMs is /etc/gwms-frontend/frontend.xml. For tar-ball installation the location is VOFRONTEND_INSTALL_DIR/frontend.xml, the canonical install dir is /var/lib/gwms-frontend/vofrontend, so you can check /var/lib/gwms-frontend/vofrontend/frontend.xml for content and ownership.  For a non standard install you could accept an install dir parameter (defaulting to /var/lib/gwms-frontend/vofrontend). You can actually use this path also for RPM installs, /var/lib/gwms-frontend/vofrontend/frontend.xml, the file content and ownership are the same as /etc/gwms-frontend/frontend.xml  (the only difference is that this copy cannot be edited directly). Startup scripts for RPMs are in /etc/init.d or /usr/sbin depending on SL version, or VOFRONTEND_INSTALL_DIR/frontend_startup for tar-ball install, so the location is even less reliable than the config script. 

I don't know when/if we'll get feedback. My proposal is to change the script to use the correct frontend user (see answer 3 above) and to include this in the next release. If admins prefer something different we'll change in future releases. What do you think?

brianhlin:
Instead of getting the install dir param to get the user/group of the config file, how about just asking for the user/group in the config (defaulting to `frontend:frontend`)?

bbockelm:
Hi - any update on this?  I had a few more pings from admins about it lately.

bbockelm:
@brianhlin - testing this out.

A few things:
1.  I like the idea of making the ownership configurable in case the `frontend` user isn't utilized.  Could be a simple config option.
2.  In my odd case, I wanted only the pilot proxy to be managed by the script, not the frontend one.  Can we make that optional?
3.  I was using the VOMS clients from CVMFS.  Can we have the `VOMS_USERCONF` environment variable override `/etc/vomses` (as that didn't exist on my host)?
4.  Also due to the fact I was using the CVMFS OSG client, I needed to source the `setup.sh` file prior to executing `gwms_renew_proxies`.  There's no clear way to do that currently, but would be nice if it was easier than writing a wrapper script and overriding the entire systemd file.  (Action here is fairly optional, as it is a "nice to have").

brianhlin:
@bbockelm I've addressed 1-3 with the latest commits

mambelli:
@brianhlin the content is OK to merge. 
This repository is only mirroring the FNAL one that requires a Kerberos credential (at least until we change the workflow). If you have a FNAL kerberos credential and would like it, I can add you to the project authors and send instructions to commit. 
If not, I can merge the patch and you should give me the exact author string you'd like me to use ('Na Me <email@host>').
For reference, this is the mirroring ticket in Redmine: https://cdcvs.fnal.gov/redmine/issues/18522

bbockelm:
I'd strongly prefer if the branch were merged as opposed to a squash commit.

brianhlin:
@mambelli I do have fnal kerberos credentials so we can go with that.

efajardo:
I am interested on the idea @bbockelm mentioned about `/etc/vomses` in the LIGO frontend we have to do something like:

`voms-proxy-init -vomses /ourown/vomses`

and the content of the `/ourown/vomses` looks like this 

`"virgo" "voms.cnaf.infn.it" "15009" "/C=IT/O=INFN/OU=Host/L=CNAF/CN=voms.cnaf.infn.it" "virgo"`

brianhlin:
@efajardo with [this commit](https://github.com/holzman/glideinWMS/pull/13/commits/1cfc622520915c25836e1468b88a1590c0879fad), you should be able to set the env var `VOMS_USERCONF to `/ourown/vomses`. If you're using the systemd file, you should be able to add an `Environment` override

brianhlin:
This was merged into `branch_v3_2`, closing.

PR #14, opened 2018-05-03
Json monitoring v2

naveenkashyap:
Pull request for a patch that generates completed_data.json to be picked up by monitoring later on.

Point of contact: Naveen Kashyap

@efajardo

### DIFF
--- a/factory/glideFactoryMonitorAggregator.py
+++ b/factory/glideFactoryMonitorAggregator.py
@@ -13,6 +13,7 @@ from __future__ import print_function
 #   Igor Sfiligoi (May 23rd 2007)
 #
 
+import json
 import copy
 import time
 import string
@@ -47,6 +48,7 @@ class MonitorAggregatorConfig:
         self.status_relname="schedd_status.xml"
         self.logsummary_relname="log_summary.xml"
         self.jobsummary_relname="job_summary.pkl"
+        self.completed_data_relname="completed_data.json"
 
     def config_factory(self, monitor_dir, entries, log):
         self.monitor_dir=monitor_dir
@@ -245,6 +247,7 @@ def aggregateStatus(in_downtime):
     global_total={'Status':None,'Requested':None,'ClientMonitor':None}
     status={'entries':{},'total':global_total}
     status_fe={'frontends':{}} #analogous to above but for frontend totals
+    completed_data_tot={'entries':{}}
 
     # initialize the RRD dictionary, so it gets created properly
     val_dict={}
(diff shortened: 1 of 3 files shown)


factory/glideFactoryMonitorAggregator.py
```diff
@@ -265,14 +268,22 @@ def aggregateStatus(in_downtime):
         # load entry status file
         status_fname=os.path.join(os.path.join(monitorAggregatorConfig.monitor_dir, 'entry_'+entry),
                                   monitorAggregatorConfig.status_relname)
+        # load entry completed data file
+        completed_data_fname=os.path.join(os.path.join(monitorAggregatorConfig.monitor_dir,'entry_'+entry),
+                                  monitorAggregatorConfig.completed_data_relname)
         try:
             entry_data=xmlParse.xmlfile2dict(status_fname)
+            completed_data_fp=open(completed_data_fname)
```
mmascher: you probably wanna add a finally clause where you close the file. Or use the with. The risk of file leaking is high otherwise.

factory/glideFactoryMonitorAggregator.py
```diff
@@ -265,14 +268,22 @@ def aggregateStatus(in_downtime):
...
+            completed_data=json.load(completed_data_fp)
         except IOError:
             continue # file not found, ignore
 
         # update entry
         status['entries'][entry]={'downtime':entry_data['downtime'], 'frontends':entry_data['frontends']}
 
+        # update completed data
+        completed_data_tot['entries'][entry]=completed_data['stats']
```
mmascher: Is the 'stat' key always present? Otherwise please use something like completed_data.get('stats', 'Unknown')

factory/glideFactoryMonitoring.py
```diff
@@ -129,6 +130,35 @@ def write_file(self, relative_fname, output_str):
...
+        Write val_dict to a json file, creating if needed
+        relative_fname: location of json relative to self.monitor_dir
+        time: typically self.updated
+        val_dict: dictionary object to be dumped to file
+        """
+
+        fname = os.path.join(self.monitor_dir, relative_fname + ".json")
+
+        try:
+                f = open(fname, 'w')
```
mmascher: Again, use with or add a finally clause

factory/glideFactoryMonitoring.py
```diff
@@ -1093,18 +1123,59 @@ def write_file(self, monitoringConfig=None):
...
+
+            completed_filename=os.path.join(monitoringConfig.monitor_dir, fe_dir) + "/Log_Completed.json"
+            completed_stats_filename=os.path.join(monitoringConfig.monitor_dir, fe_dir) + "/Log_Completed_Stats.json"
+            completed_wastetime_filename=os.path.join(monitoringConfig.monitor_dir, fe_dir) + "/Log_Completed_WasteTime.json"
+
+            try:
+
+                completed_fp = open(completed_filename)
+                completed_stats_fp = open(completed_stats_filename)
+                completed_wastetime_fp = open(completed_wastetime_filename)
```
mmascher: Again, please make sure to use the correct ways to handle files

factory/glideFactoryMonitorAggregator.py
naveenkashyap: The 'stats' key will always be present. When the completed_data.json file is written, the data is written to the 'stats' key. However, this does rely on no one changing how completed_data.json is written. Should I make your suggested change for added robustness?

factory/glideFactoryMonitoring.py
naveenkashyap: Done

factory/glideFactoryMonitorAggregator.py
naveenkashyap: Done

naveenkashyap:
The latest commits implement a jsonMonitor that uses the completed_data.json created in previous commits to aggregate schedd_status and completed_data and store it in influxdb to be picked up by grafana downstream. 

Note: the monitor does not work out of the box. the config file in jsonMonitoring/config/config.json needs to be configured. Also, the spec file (for RPM installation/upgrades) needs to be updated to place the logrotate file (jsonMonitoring/logrotate/factory-monitor) in the appropriate location.

mambelli:
Accepted and merged in v3.4
See https://cdcvs.fnal.gov/redmine/issues/20018

PR #15, opened 2018-06-20
restore old color scheme for factoryStatus.html but updated for core …

jdost321:
…versions


mambelli:
This branch was included in v3_4_1
See ticket https://cdcvs.fnal.gov/redmine/issues/20295

PR #21, opened 2019-06-20
Set WN in Drain only if less than 4h is left.

juztas:
Some discussion here: https://its.cern.ch/jira/projects/CMSSI/issues/CMSSI-13?filter=allopenissues

@mmascher


juztas:
@mmascher any news on this?

mambelli:
Rejecting the changes.
This was solved with a more complete solution in  https://cdcvs.fnal.gov/redmine/issues/22867

juztas:
Checked one of the latest pilots and this is not in place. Is it in production?

PR #24, opened 2019-11-12
Update proxy renewal make rfc default.

aaarora:
Added '-rfc' to CMD in voms_proxy_fake()

### DIFF
--- a/frontend/gwms_renew_proxies.py
+++ b/frontend/gwms_renew_proxies.py
@@ -179,7 +179,8 @@ def voms_proxy_fake(proxy, vo_info):
            '-hostcert', vo_info.cert,
            '-hostkey', vo_info.key,
            '-uri', vo_info.uri,
-           '-fqan', vo_info.fqan]
+           '-fqan', vo_info.fqan
+           '-rfc']
     return _run_command(cmd)
 
 


efajardo:
@mambelli can you review this. All proxies should come as RFC anyway but this is definetly a bug

efajardo:
@brianhlin I am tagging you since you developed this. @aaarora is a student of mine working on containarizing GWMS parts

brianhlin:
I always forget whether we want RFC or not... I seem to remember @rynge knowing more about this but I could be completely wrong...

efajardo:
we always want RFC

rynge:
We use legacy proxies for OSG VO - but probably because that is what gwms proxy script creates.

At some point we had issues with RFC ones on some European CEs.

brianhlin:
@efajardo it sounds like then we need to add this as an option to the configuration so that operators can decide which proxies are generated as RFC vs non-RFC, defaulting to the latter.

brianhlin:
@rynge perhaps we should go with RFC as the default and lean on those European sites to accept RFC proxies?

rynge:
Sure, I have no preference.

brianhlin:
In that case, @efajardo, I'm on board with this change

efajardo:
@mambelli so can we merge this?

mambelli:
The Fermilab one is still the main repo. Changes are in b3fd8c1aa7d1d0b53e56029ef63928b3c21b5b49
I added also a comma that was missing.

efajardo:
@mambelli in fact there is a mistake here. If we use `-rfc` we also have to use `-pathlength 20` or some big number.

PR #25, opened 2020-02-11
Adding GPU monitor as default for pilots

efajardo:
The GPU monitor was fixed by HTCOndor https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=6857 so we should be deplying this by default


mambelli:
I made a ticket to handle this and there is a solution that is better than the one proposed here. In this pull request, the code sets up GPUs, attributes, and monitoring also when the user requests something different form GPUs. The plan is to go with the code in branch v36/24112 and reject this PR. See https://cdcvs.fnal.gov/redmine/issues/24112 for more

mambelli:
Closing. The code in v36/24112 solving the PR is already in master (see the previous comment)

PR #26, opened 2020-04-22
Fix RFC Proxy Regenration

efajardo:


efajardo:
@brianhlin could you please review this?

frontend/gwms_renew_proxies.py
```diff
@@ -88,6 +92,8 @@ def __init__(self, vo, fqan):
         self.name = vo
         if fqan.startswith('/%s/Role=' % vo):
             pass
+        elif fqan.startswith('/%s/' % vo):
```
brianhlin, line 95: Instead of adding an `elif` here, let's just loosen the restriction in the initial if to:

```
if fqan.startswith('/%s/' % vo):
```

frontend/gwms_renew_proxies.py
```diff
@@ -213,9 +221,18 @@ def main():
...
         proxy_config = dict(config.items(proxy_section))
+        if 'rfc' not in proxy_config:
+            proxy_config['rfc'] = True
+        else:
+            if safe_boolcomp(proxy_config['rfc'], False):
+                proxy_config['rfc'] = False
+            else:
+                proxy_config['rfc'] = True
+        if 'path_length' not in proxy_config:
+             proxy_config['path_length'] = '20'
```
brianhlin, line 232: I don't think you need this block new block because we load the defaults here:

https://github.com/glideinWMS/glideinwms/blob/3eb90b1901924bf25c75f9b486bc7aeddd2bd0cc/frontend/gwms_renew_proxies.py#L198-L200

frontend/gwms_renew_proxies.py
```diff
@@ -22,6 +22,8 @@
             'fqan': '/Role=NULL/Capability=NULL',
             'frequency': '1',
             'lifetime': '24',
+            'path-length': '20',
```
brianhlin, line 25:
```suggestion
            'path_length': '20',
```

efajardo:
Thanks for reviewing @brianhlin I applied all your changes. @mambelli I tested this and It is working. So I will let you decide how to merge.

mambelli:
Merged manually to the Fermilab Git repo.

PR #30, opened 2020-08-04
Remove quotes from singularity binary path

efajardo:


mambelli:
I don't think the removal of the quotes from the variable is changing anything.
It is in the shell assignment and although it is not required normally in bash (would be in sh/POSIX) it is recommended and it protects from edge cases (the value is not a straight forward right value).

I think the problem causing the quotes condor value is in the advertise command in singularity_setup.sh where SINGULARITY_PATH and GWMS_SINGULARITY_PATH are considered strings.
I'm not sure if file paths in the condor config macros should be considered strings or expressions (i.e. if they require quotes or not).

mambelli:
Closing rejecting the changes.
The quotes are actually needed. Without there is an HTCondor configuration error, see https://cdcvs.fnal.gov/redmine/issues/24953

PR #32, opened 2020-10-16
Add the ability to set the X509 proxies strength

efajardo:


efajardo:
@brianhlin and @mambelli this is needed. For RHEL8 the condor daemons need to talk to each other using proxies of 2048 bits otherwise they complain. I am leaving the default as `1024` strength as it is today and I will be testing in some frontends over the weekend what happens if the default is moved to 2048.

brianhlin:
You should also add this to the default config along with a comment

efajardo:
thanks @brianhlin I add it to the default config

PR #39, opened 2021-01-14
Disable LD_PRELOAD during container execution

rynge:


mambelli:
Created ticket https://cdcvs.fnal.gov/redmine/issues/25428 and applied the same changes also to singularity_lib.sh (prepare_and_invoke_singularity in the wrapper is being replaced with a function there - note in the comments)

PR #47, opened 2021-04-16
Modify gmws-fronted web configuration to find images

aaronmoate:
This is related to Jira ticket SOFTWARE-4578
https://opensciencegrid.atlassian.net/browse/SOFTWARE-4578

### DIFF
--- a/build/packaging/rpm/gwms-frontend.conf.httpd
+++ b/build/packaging/rpm/gwms-frontend.conf.httpd
@@ -1,8 +1,21 @@
 # This is the httpd conf file
 # GlideinWMS VOFrontend web configuration
 
+Alias /vofrontend/monitor/images /var/lib/gwms-frontend/web-base/frontend/images/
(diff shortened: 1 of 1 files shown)


mambelli:
@aaronmoate The files should be also in `/var/lib/gwms-frontend/web-area/monitor/images` (line 428 and 836, in glideinwms-vofrontend-core), which is within the web space/Alias already defined.
The files are also in web-base to allow the `upgrade` command but should not be served by that directory.

mambelli:
I verified all current releases and the sources (spec files) and all do include the files in the right place. Somehow the installation missing `/var/lib/gwms-frontend/web-area/monitor/images` was corrupted. `rpm -V glideinwms-vofrontend-core` should detect the missing files. I don't know how that happened, so I'm leaving issues/36 open but this proposed solution is incorrect, so I'm closing the PR without merging.

PR #48, opened 2021-04-17
Fix IDTOKENS in the frontend.

bbockelm:
This fixes two things:
1. Removes a condor version check that is non-functional.
2. Name the idtoken parameter to be identical to the EntryName; this name is required by the factory.


ddbox:
Reviewed and tested just to be on the safe side.  Merged on the Redmine side from FNAL

ddbox:
Reviewed and tested just to be on the safe side.  Merged on the Redmine side from FNAL

PR #51, opened 2021-04-30
Have SciToken credential be per-entry

bbockelm:
The CHTC frontend generates a unique token per entry point.  This prevents a token on CE1 from being stolen and submitting to CE2.

However, the factory internally stores them all under the same filename.  With this change, the credential is stored per-entry.


ddbox:
this is actually redmine issue #25934.  Branch v37/25934 implementing this was merged to master but the pull request was never updated

PR #52, opened 2021-04-30
Have SciTokens generated by an external script.

bbockelm:
Instead of assuming that a SciToken is in an external directory (which requires the admin to know all the possible GLIDEIN_Sites that the frontend wants to submit to in advance), use the same logic as found in the IDTOKEN case: have the frontend invoke an admin-provided script.

Open items:
* Currently, there's too much variation in how tokens are generated: is it reasonable to provide a default script?  For now, I provide nothing.
* This stores the credential in a per-frontend directory.  This really should be stored in a per-group directory (and perhaps the group name is an input argument to the script?).


ddbox:
moving this to a feature request in redmine.
https://cdcvs.fnal.gov/redmine/issues/25985

PR #72, opened 2022-02-11
Some changes to make manual_glidein_submit python3 compatible

mmascher:


factory/tools/manual_glidein_submit.py
```diff
@@ -152,7 +152,8 @@ def main():
             ad_gc, frontend_descript, glidein_descript.data['PubKeyObj'])
         security_class = sym_key_obj.decrypt_hex(
             ad_gc['GlideinEncParamSecurityClass']) # GlideinSecurityClass
-        proxyid = sym_key_obj.decrypt_hex(ad_gc['GlideinEncParamSubmitProxy'])
+        proxyid = sym_key_obj.decrypt_hex(ad_gc['GlideinEncParamSubmitProxy']).decode('utf8')
+        security_class = security_class.decode('utf8')
```
mambelli, line 156: You could add `.decode('utf8')` at the end of line 154, to be consistent w/ proxyid

mambelli (changes requested):

factory/tools/manual_glidein_submit.py
mmascher: Done

mambelli (approved):

PR #75, opened 2022-02-14
[v3_7_6] Do not clear LD_LIBRARY_PATH before starting condor. Fixes #73

### LINKED ISSUE #73: Revert unsetting of LD_LIBRARY_PATH before starting condor
GRIF_IRFU had a local installation of HTCondor which was the same of the one used by CMS for the pilot tarballs. Since they had LD_LIBRARY_PATH pointing to it, there were issues because the pilot tarballs and the local installation were for different OS versions. See: https://ggus.eu/index.php?mode=ticket_info&ticket_id=154983

We decided to unset LD_LIBRARY path right before starting the condor_master, with the assumption the condor tarball is self contained and does not need anything. However, with 3.7.6 this caused other type of issues. For example, at CSCS libcrypto was not found (it's a Suse OS IIRC https://ggus.eu/index.php?mode=ticket_info&ticket_id=155698):

```
/scratch/lcg/daint/arc_session/arc05_1/9vxNDmGuHX0nBq2IepesqxioABFKDmABFKDmk5KKDmJBFKDmi3nlVm/glide_cwfznX/main/condor/sbin/condor_advertise: error while loading shared libraries: libcrypto.so.10: cannot open shared object file: No such file or director
```

And at other sites there were issues with finding cuda for the gpu discovery tool:

```
Can't open libcuda.so. Error 19587456
DetectedGPUs=0
```

I have patched the production instances and removed the unset.

mmascher:


mambelli (approved):

mmascher:
@mambelli This was supposed to be the PR for the 3.7 branch. Did I miss to point to t hat branch?

PR #80, opened 2022-02-16
Make sure submit attributes are set in the OSG.yaml file. Fixes #79

### LINKED ISSUE #79: Make sure submit attributes are set in the OSG.yaml file
That's a change was lost during resolution of conflicts when a previous change was merged. Interestingly, unit tests caught this but I thought it was a false positive anche changed the test! I'll revert the unit test change as well.

mmascher:


mmascher:
The unit test failure is a legitimate bug:

```
expected_out['LSU']['hosted-ce29.grid.uchicago.edu']['WholeNode']['DEFAULT_ENTRY']['submit_attrs']
(Pdb) {'+WantWholeNode': True}
get_information_internal(info)['LSU']['hosted-ce29.grid.uchicago.edu']['WholeNode']['DEFAULT_ENTRY']['submit_attrs']
(Pdb) {'+xcount': '', '+WantWholeNode': True}
```

The entry is a wholenode one, so it should not have xcount, not even empty.

mmascher:
The bug was introduced because we changed the default for CPUs here https://github.com/glideinWMS/glideinwms/commit/7ac93d875c9648ba19233f5043e4653703d01b3a#diff-81f7c107733ea964122b1c06682cae166ca9dbb84e00438882d4c31b579b7e64L68

I fixed this by changing the defaults to None when we get the values from the collector.

PR #111, opened 2022-04-07
Add line terminator to the end of decoded SciToken

BrunoCoimbra:
This PR fixes an issue causing SciTokens, transferred from the Frontend, to be written on the Factory without line terminators. This problem would cause the resulting SciToken to be impossible to decode.

### DIFF
--- a/factory/glideFactoryEntry.py
+++ b/factory/glideFactoryEntry.py
@@ -1278,7 +1278,6 @@ def unit_work_v3(
     scitoken_file = os.path.join(submit_credentials.cred_dir, scitoken)
     scitoken_data = decrypted_params.get("frontend_scitoken")
     if scitoken_data:
-        scitoken_data = scitoken_data.rstrip("\n ")
         if token_util.token_str_expired(scitoken_data):
             entry.log.warning(
                 "frontend_scitoken supplied by frontend, but expired. Renaming to %s.expired" % scitoken_file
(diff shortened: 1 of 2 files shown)


factory/glideFactoryEntry.py
```diff
@@ -1292,7 +1292,7 @@ def unit_work_v3(
             try:
                 entry.log.info("frontend_scitoken supplied, writing to %s" % scitoken_file)
                 chmod(tmpnm, 0o600)
-                os.write(fd, scitoken_data.encode("utf-8"))
+                os.write(fd, (scitoken_data + os.linesep).encode("utf-8"))
                 os.close(fd)
                 util.file_tmp2final(scitoken_file, tmpnm)
```
mambelli, line 1297: Had a bunch of comments but clicked the wrong button and loast all. Mostly about pre-existing code but I think there is alot of back and forth that can be spared:
1. Maybe the token data should be written as received from the decrypted classad w/o many changes, and a second string used locally for checks, maybe adding the strip() in `token_util.token_str_expired`
2. if stripping the end do also the beginnig. I don't think tokens expect spaces/newlines at the beginning, `scitoken_data.strip("\n ")` . See 1 about doing it in the verify
3. What about instead of os.write, os.close:
```
with fdopen(fd, 'w', encoding="utf-8") as f:
    f.write(f"{scitoken_data}\n")
```
or `print(scitoken_data, file=f)`, both use universal newline (whatever the os wants except when forced in the file opening. close is automatic (also for fd), this defaults to text while mkstemp defaults to binary

mambelli (changes requested):

lib/token_util.py
```diff
@@ -85,7 +85,8 @@ def token_str_expired(token_str):
     expired = True
     try:
         decoded = jwt.decode(
-            token_str, options={"verify_signature": False, "verify_aud": False, "verify_exp": True, "verify_nbf": True}
+            token_str.strip(),
```
mambelli, line 88: Use `token_str.strip("\n ")` to protect also against spaces like it was before

factory/glideFactoryEntry.py
```diff
@@ -1288,12 +1287,12 @@ def unit_work_v3(
             if "frontend_scitoken" in submit_credentials.identity_credentials:
                 del submit_credentials.identity_credentials["frontend_scitoken"]
         else:
-            (fd, tmpnm) = tempfile.mkstemp(dir=submit_credentials.cred_dir)
             try:
                 entry.log.info("frontend_scitoken supplied, writing to %s" % scitoken_file)
+                (fd, tmpnm) = tempfile.mkstemp(dir=submit_credentials.cred_dir)
+                with os.fdopen(fd, "w", encoding="utf-8") as f:
+                    f.write(f"{scitoken_data}\n")
```
mambelli, line 1294: Do we still need to add the newline now that we don't strip the data from the classad?

mambelli (changes requested):

lib/token_util.py
mambelli: Bruno assured that this is the default behavior, space and newlines are both stripped

mambelli (approved):

Issue #63, opened 2021-12-17
Add attribute for singularity variant


matyasselmeci:
Now that there are multiple forks of Singularity (e.g. AppTainer, Singularity CE), in addition to `SINGULARITY_VERSION`, it would be useful to add the variant as an attribute (suggested name: `SINGULARITY_VARIANT`). According to @DrDaveD, this is the first word of the output of `singularity --version`, for example
```
singulary-ce version 3.9.0-rc.3
```
should result in `SINGULARITY_VARIANT="singularity-ce"`.

This is especially important since AppTainer plans to reset their version to 1.0.0 and start numbering from there.

mambelli:
@matyasselmeci  Have a counter-proposal
What about `SINGULARITY_FULL_VERSION`, containing the whole string returned by `singularity --version`
Or, even better `CONTAINERSW_FULL_VERSION`, since it is not only singularity. It is also apptainer, soon maybe also podman.

In short, my preference would be:
1. CONTAINERSW_FULL_VERSION containing the whole version string
2. SINGULARITY_FULL_VERSION containing the whole version string (if 1 is not OK)

matyasselmeci:
I'm OK with using the whole version string. The issue with CONTAINERSW_FULL_VERSION though is what if you have multiple runtimes installed, e.g. both singularity and podman - which one would you report?

mambelli:
The only one used by the glidein. 
Currently it is a flavor of singularity, the first one encountered in the search places, with preference for apptainer, then singularity, then singularity-pro (singularity-ce). 
GlideinWMS is advertising also:
``` 
CONTAINERSW_PATH
CONTAINERSW_VERSION
CONTAINERSW_FULL_VERSION
CONTAINERSW_MODE
```
For now, these are the same as the SINGULARITY_ variables. The idea is that in the future someone will require also the support of another container software and these variables will all point to the one that has to be used (as requested by the VO or the site). Probably a CONTAINERSW_NAME or CONTAINERWS_ID will have to be added to select different behaviors depending on the name of the container sw. I'd like the idea to start thinking along these ways because there is no guarantee that apptainer will not diverge in behavior from singularity

mambelli:
Do you think in the future we'll have to advertise the availability and versions of multiple container software (all the ones that are installed and available)?
In this case, we may need a `SINGULARITY_FULL_VERSION`, `PODMAN_FULL_VERSION` and maybe also `APPTAINER_FULL_VERSION`.

[...]

PR #118, opened 2022-04-14
Advertising CONTAINERSW_FULL_VERSION and CONTAINERSW_ variables

mambelli:
Advertising CONTAINERSW_ variables, especially CONTAINERSW_FULL_VERSION that will be used by OSG
It is a first step in the generalization of container software support
This is solving issue #63

### DIFF
--- a/creation/web_base/singularity_lib.sh
+++ b/creation/web_base/singularity_lib.sh
@@ -1504,6 +1504,10 @@ singularity_locate_bin() {
     export GWMS_SINGULARITY_PATH=""
     export GWMS_SINGULARITY_VERSION=""
     export GWMS_SINGULARITY_MODE=""
+    export GWMS_CONTAINERSW_PATH=""
+    export GWMS_CONTAINERSW_VERSION=""
+    export GWMS_CONTAINERSW_FULL_VERSION=""
+    export GWMS_CONTAINERSW_MODE=""
     warn "Singularity not found$s_location_msg in OSG_SINGULARITY_BINARY[_DEFAULT], PATH and module"
     info_dbg "PATH(${PATH}), attempt results(${bread_crumbs})"
     false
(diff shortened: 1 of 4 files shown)


creation/web_base/singularity_lib.sh
```diff
@@ -2106,6 +2113,10 @@ setup_classad_variables() {
     export GWMS_SINGULARITY_STATUS=$(get_prop_str ${_CONDOR_MACHINE_AD} GWMS_SINGULARITY_STATUS)
     export GWMS_SINGULARITY_PATH=$(get_prop_str ${_CONDOR_MACHINE_AD} SINGULARITY_PATH)
     export GWMS_SINGULARITY_VERSION=$(get_prop_str ${_CONDOR_MACHINE_AD} SINGULARITY_VERSION)
+    export GWMS_CONTAINERSW_STATUS=$(get_prop_str ${_CONDOR_MACHINE_AD} GWMS_CONTAINERSW_STATUS)
```
namrathaurs, line 2116: I do not see any reference to the variable `GWMS_CONTAINERSW_STATUS`, used in the last part of this line, in any of the modified files. Just making sure that this does not cause any issues when in execution. 

I do see a variable named `GWMS_CONTAINERSW_MODE` but is not used anywhere in the code.

namrathaurs (changes requested):

creation/web_base/singularity_lib.sh
mambelli: These variables are there to try to future proof. Currently there is only one container sw (singularity in its variants).

I removed GWMS_CONTAINERSW_STATUS. Thinking better it does not make sense yet since the container sw selection may use different mechanisms. Maybe in the future when things are more defined.

GWMS_CONTAINERSW_MODE/CONTAINERSW_MODE are staying OSG and others outside GWMS may start using it.

I pushed an update

PR #122, opened 2022-04-20
Porting Redmine tickets for on-demand CVMFS to v3.7.7

namrathaurs:
(26094) Automating the creation of platform-specific cvmfs distribution
- Instead of shipping a large tarball with all possible combinations of distributions, one per platform architecture, evaluate the worker node to determine the system information and dynamically create a cvmfs distribution for the underlying platform and architecture.

(26153) Version-based download/build of cvmfsexec distribution tar files
- Enable download/build of cvmfsexec distributions only when there is a version difference

(26713) Include on-demand CVMFS provisioning info in GWMS documentation
- Added documentation regarding on-demand CVMFS (the internal specifics) and how the dynamic creation/selection of the cvmfsexec distribution takes place.

PR #123, opened 2022-04-20
Porting Redmine ticket 26153 to v3.7.7

namrathaurs:
Version-based download/build of cvmfsexec distribution tar files 

- enable download/build of cvmfsexec distributions only when there is a version difference


namrathaurs:
Abandoning and closing this PR since a single PR has been created (#122) that includes this ticket to avoid conflicts.

PR #131, opened 2022-04-27
Add the possibility to configure the lifetime of idtokens. Fixes #129

### LINKED ISSUE #129: Add the possibility to configure the lifetime of idtokens
IDtokens used to connect WN to VO collectos are hardcoded at 24 hours of lifetime. This means that if the pilot stays idle for more than 24 hours in the CE then the pilot's startd won't be able to connect back to the collector. Startd disconnections are also an issue if the disconnection happens after 24hours.

Add a parameter in the frontend to configure the idtoken lifetime

A refreshing mechanism that sends the idtoken from the factory to the CE then the batch system head node and finally the WN would need condor support and is not on the horizon.

mmascher:


ddbox (approved):
looks good

PR #134, opened 2022-04-27
Issue 128 branch v3 7

ddbox:
Frontend operator adds
"""
&lt; attr name="CONTINUE_IF_NO_PROXY" glidein_publish="True" job_publish="True" parameter="True" type="string" value="True"/ &gt;
"""
to a frontend group if it has a scitoken credential only to enable routing of credentials in factory.
If parameter is missing or has value other than 'True' factory behaves as it did in 3.7.6


mambelli (approved):

mambelli:
OK to merge once the unit test is fixed and this branch rebased w/ no unit test errors

Issue #152, opened 2022-06-09
Frontend talking to multiple Factories w/ same entry name causes token generation collision and inconsistencies


mambelli:
**Describe the bug**
If a Frontend talks to multiple Factories and they have the same entry name, this will cause a collision in the token generation plugin

The current API uses:
```
log
group (str): Frontend group
entry (dict): Factory entry information dictionary, containing at least:
    name (str): the entry name, and
    gatekeeper (str): the gatekeeper string
trust_domain (str): Credential trust domain
tkn_dir (str, optional): Directory where the tokens are stored. Defaults to "/var/lib/gwms-frontend/tokens.d".
```
As you can see there is no factory. So the plugin will use the same token if 2 Factories have an entry w/ the same name. 

So if some other information like the one in the audience is different generated tokens are inconsistent.

**To Reproduce**
Have a frontend talk to 2 Factories, with a common entry name and generate scitokens for that entry

**Expected behavior**
There should be no conflict. Different factories could use the same entry name and these should be considered different entries.
Tokens for an entry in a Factory should be consistent.

**Screenshots**
If applicable, add screenshots and/or console outputs to help explain your problem.

**Desktop (please complete the following information):**

-   GlideinWMS version: 3.9.5
-   Python version: 3.6
-   OS version: SL7
-   HTCondor version: any

**Additional context**
Add any other context or supporting files about the problem here.

PR #157, opened 2022-06-15
Prevent collision when generating credentials

BrunoCoimbra:
This pull request fixes issue #152.


mambelli (approved):
There were a couple of small fixes (docsting and unnecessary chmode) that I added to the PR

PR #160, opened 2022-06-20
Fix exception duting completed stats computation. Fixes #155

### LINKED ISSUE #155: Completed stats are missing after 3.9.5 upgraded
One of the monitoring pages was showing condor failing to start and triggering alarm for the UCSD factory after the upgrade. It turns out that the `rrd_Log_Completed.xml` file was not containing correct information. After some investigation I found out this was the cause of the issue:

```
>>> from glideinwms.factory.glideFactoryLogParser import extractLogData
>>> extractLogData("/var/log/gwms-factory/client/user_fecmsglobal/glidein_gfactory_instance/entry_CMSHTPC_T2_AT_Vienna_ce-1/job.10034598.7.out")
  File "/usr/lib/python3.6/site-packages/glideinwms/factory/glideFactoryLogParser.py", line 376, in extractLogData
    validate_re = ELD_RC_VALIDATE_END.search(buf, buf_idx)
TypeError: cannot use a string pattern on a bytes-like object
```

I fixed it by adding a `.read().decode("utf-8")` to the buffer. It seems to work without adding any cost to the performances. Still keeping an eye on it.

mmascher:


mambelli (changes requested):
I looked into the whole file to understand better the use.
This is used to parse the glidein stdout, which may be fairly big. And there are many glideins.
I don't think reading it all is a good solution, the use of memory and time may explode.

I suggest the opposite approach. Start with the original version. The file is a bytes file, convert the regular expressions to bytes, and convert (decode) only the outputs to str (few are strings, others are numbers which are OK, int(bytes:) works fine)

mmascher:
I was also worried about performances, but I did not see any change in the memory usage charts:https://snapshots-origin.grafana.net/dashboard/snapshot/zYf6xB5iiKTP4st67i2qAcsbvVuWa5E6?orgId=2&refresh=1m
I installed the factory on the 14th/15th of June and applied the fix right after.

Anyway, the changes you are proposing makes a lot of sense! I did not think of converting the regexp to bytes.

mmascher:
I opened a new PR since the changes here do not make sense anymore: https://github.com/glideinWMS/glideinwms/pull/165
Should we close this one and start over from the new one?

PR #161, opened 2022-06-21
Adding GLIDEIN_PERIODIC_SCRIPT to env to notify user scripts

mambelli:
Adding GLIDEIN_PERIODIC_SCRIPT to env to notify a user script that is… invoked under startd_cron, periodically, and not at startup

Added also linting suggestions to the shell script


namrathaurs (approved):
Changes look fine. Ready to merge.

Issue #164, opened 2022-06-21
setup_x509.sh can write to glidein_config as a startd cron


brianhlin:
The first comments in the main section of `setup_x509.sh` indicates that it should only try to refresh the proxy on its first run:

https://github.com/glideinWMS/glideinwms/blob/master/creation/web_base/setup_x509.sh#L379-L380

However, it will perform its initial setup steps if it doesn't succeed in refreshing an X.509 proxy or token. Combined with #163, this can result in malformed `glidein_config` files

PR #165, opened 2022-06-24
Fix exception duting completed stats computation. Fixes #155

### LINKED ISSUE #155: Completed stats are missing after 3.9.5 upgraded

mmascher:


mmascher:
I see we have a dir named `unittests/fixtures/factory/log/client/` in the repo. It is empty. Is it supposed to contain glidein logfiles that can be used for unit tests? I was thinking of adding a real(production) glidein stdout logfile and adding a unit test.

mmascher:
I just closed https://github.com/glideinWMS/glideinwms/pull/164 that preceeded this one.

mambelli:
Changers satisfy the suggestions in PR #160.
`unittests/fixtures/factory/log/client/` and a subdirectory under it are the right directory for the log file as discussed in the meeting today. Waiting for the unit test before approving and merging.

mmascher:
@mambelli I run the test on my machine and I could not replicate the issue. Could it be related to a different version of the system `m2crypto` library? I have '0.35.2' on my machine.

I see the output of the unit test is saved here `output/gwms.LOCAL.pyunittest.d/gwms.LOCAL.pyunittest.d.test_lib_symCrypto.txt` but I could not find out how to access it. Can you help me? :)

PR #166, opened 2022-06-28
make default glidein_config robust to parallel updates

mambelli:
Made glidein_config more robust to parallel updates.
It will never be inconsistent but can still miss (not record) some updates done in parallel.
This is a transitional step, the next one, to make the process even more robust, will change the glidien_config format so it needs all the custom scripts to use the provided functions to save and retrieve attributes.

The documentation and the project changelog have been updated as well

### DIFF
--- a/creation/web_base/add_config_line.source
+++ b/creation/web_base/add_config_line.source
@@ -7,30 +7,121 @@ warn() {
  echo "$(date)" "$@" 1>&2
 }
 
+# Mac compatibility
+TAC=$(command -v tac)
+[[ -z "$TAC" && "$OSTYPE" =~ darwin* ]] && TAC="tail -r"
+
+# glidein_config is a key-value file
+# each line is a key (case sensitive ID name), a space, and a value (rest of the line)
+# if there are multiple lines w/ the same key, the last one counts
+# ecah linke can be max 4K
+GWSM_CONFIG_LINE_MAX=4k
+
+# Return the value stored in glidein_config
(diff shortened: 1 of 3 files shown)


creation/web_base/add_config_line.source
```diff
@@ -7,30 +7,85 @@ warn() {
  echo "$(date)" "$@" 1>&2
 }
 
+# return one value stored in glidein_config
+# Arg: id (key) of the value to retrieve
+get_config() {
+    grep ""^$1 " "$glidein_config" | cut -d ' ' -f 2-
```
matyasselmeci, line 13:
```suggestion
    grep "^$1 " "$glidein_config" | cut -d ' ' -f 2-
```

mambelli: Fixed that when I changed the function

creation/web_base/add_config_line.source
```diff
@@ -7,30 +7,121 @@ warn() {
  echo "$(date)" "$@" 1>&2
 }
 
+# Mac compatibility
+TAC=$(command -v tac)
+[[ -z "$TAC" && "$OSTYPE" =~ darwin* ]] && TAC="tsil -r"
```
matyasselmeci, line 12:
```suggestion
[[ -z "$TAC" && "$OSTYPE" =~ darwin* ]] && TAC="tail -r"
```

AidanHimley (approved):

PR #170, opened 2022-07-01
Allow Frontend schedds not to be in DNS

mambelli:
Failing Frontend configuration validation (reconfig/upgrade) only if all HTCSS schedds are unknown to DNS

Especially in production, there may be many VO HTCSS schedds, some not regularly maintained (coming up and down also from DNS)

### DIFF
--- a/creation/lib/cvWParamDict.py
+++ b/creation/lib/cvWParamDict.py
@@ -32,7 +32,7 @@ from .matchPolicy import MatchPolicy
 
 ####################################################################################
 # Functions to validate the match expression once expanded
-# validate_node() was removed, should these also be removed or moved somewhere else?
+# validate_node() was moved to cWDictFile, should these also be removed or moved somewhere else?
 
 
 def translate_match_attrs(loc_str, match_attrs_name, match_attrs):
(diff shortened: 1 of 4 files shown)


creation/lib/cvWParamDict.py
```diff
@@ -1100,9 +1097,21 @@ def populate_common_descript(descript_dict, params):
...
+        # Would be useful to have a WARNING message, but the current implementation allows only fail/continue
+        # Still raising an invalid configuration exception if no schedd is in DNS
+        try:
+            cWDictFile.validate_node(el["fullname"], check_dns=False)
+            valid_schedd = True  # skipped if exception is risen
+        except RuntimeWarning:
+            undefined_schedds += 1
         schedds.append(el["fullname"])
+    if undefined_schedds > 0 and not valid_schedd:
+        raise RuntimeError("No valid schedd found, all are unknown to DNS")
```
BrunoCoimbra, line 1114: Having `valid_schedd` and `undefined_schedds` here seems redundant to me. If we don't need the exact number of undefined schedds, `valid_schedd` seems to be enough.

BrunoCoimbra (approved):

PR #178, opened 2022-07-09
Fix fair_split and support token proxy pairs

mambelli:
Fixed splitting formula using integer division (like in the pythin2 version)
Added workaround for the x509+token special case


BrunoCoimbra:
The issues that I was still having with the DE were related to the `CONTINUE_IF_NO_PROXY` variable. The changes in the PR should be ready to go.

BrunoCoimbra (approved):

PR #180, opened 2022-07-13
Removed the need for X509 credential (Issue 66)

### LINKED ISSUE #66: Frontend group throws exception unless X509 credential is provided
- If a frontend group doesn't have a credential assigned, then it will throw an exception and fail.
- If you fix the obvious exceptions, you quickly run into a problem with the frontend/factory interface: the request advertisements are per-credential.  As the token implementations don't utilize the credential framework, there's no way for the frontend to generate the ad names to advertise to the factory.

ddbox:
Fixes #66

### DIFF
--- a/creation/lib/cvWCreate.py
+++ b/creation/lib/cvWCreate.py
@@ -61,6 +61,7 @@ def create_client_mapfile(mapfile_fname, my_DN, factory_DNs, schedd_DNs, collect
     """
     with open(mapfile_fname, "w") as fd:
         try:
+            # issue 66 transition from GSI, keep going if no DNs for proxies
             fd.write('GSI "^{}$" {}\n'.format(re.escape(my_DN), "me"))
         except:
             pass
--- a/creation/lib/cvWCreate.py
+++ b/creation/lib/cvWCreate.py
@@ -60,7 +60,10 @@ def create_client_mapfile(mapfile_fname, my_DN, factory_DNs, schedd_DNs, collect
 
     """
     with open(mapfile_fname, "w") as fd:
-        fd.write('GSI "^{}$" {}\n'.format(re.escape(my_DN), "me"))
+        try:
+            fd.write('GSI "^{}$" {}\n'.format(re.escape(my_DN), "me"))
+        except:
+            pass
         for (uid, dns) in (
             ("factory", factory_DNs),
             ("schedd", schedd_DNs),
--- a/creation/lib/cvWCreate.py
+++ b/creation/lib/cvWCreate.py
@@ -60,11 +60,8 @@ def create_client_mapfile(mapfile_fname, my_DN, factory_DNs, schedd_DNs, collect
 
     """
     with open(mapfile_fname, "w") as fd:
-        try:
-            # issue 66 transition from GSI, keep going if no DNs for proxies
+        if my_DN:
             fd.write('GSI "^{}$" {}\n'.format(re.escape(my_DN), "me"))
-        except:
-            pass
         for (uid, dns) in (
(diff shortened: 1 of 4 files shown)


brianhlin:
This is for #66

creation/lib/cvWParamDict.py
```diff
@@ -1273,8 +1277,8 @@ def populate_gridmap(params, gridmap_dict):
 #####################################################
 # Populate security values
 def populate_main_security(client_security, params):
-    if params.security.proxy_DN is None:
-        raise RuntimeError("DN not defined for classad_proxy")
+    # if params.security.proxy_DN is None:
+    #    raise RuntimeError("DN not defined for classad_proxy")
     client_security["proxy_DN"] = params.security.proxy_DN
```
mambelli, line 1282: Probably the assignment should be conditional to the proxy being there.
There may be some compensation downstream needed if `client_security["proxy_DN"]` is empty. 
Check if it should do None or "".

creation/lib/cvWParamDict.py
```diff
@@ -851,11 +851,15 @@ def populate_frontend_descript(work_dir, frontend_dict, active_sub_list, params)
...
-        raise RuntimeError("security.classad_proxy(%s) is not a file" % params.security.classad_proxy)
+    # for now, just allow absence of proxies in face of JWT hedgemony
+    # if params.security.classad_proxy is None:
+    #    raise RuntimeError("Missing security.classad_proxy")
+    try:
+        params.subparams.data["security"]["classad_proxy"] = os.path.abspath(params.security.classad_proxy)
+    except:
+        params.subparams.data["security"]["classad_proxy"] = None
+    # if not os.path.isfile(params.security.classad_proxy):
+    #    raise RuntimeError("security.classad_proxy(%s) is not a file" % params.security.classad_proxy)
```
mambelli, line 862: Try to avoid generic `except`, if there is a name and the file is not there, it is still an error. Something like:
```
    if params.security.classad_proxy is None:
        params.subparams.data["security"]["classad_proxy"] = None
    else:
        params.subparams.data["security"]["classad_proxy"] = os.path.abspath(params.security.classad_proxy)
        if not os.path.isfile(params.security.classad_proxy):
            raise RuntimeError("security.classad_proxy(%s) is not a file" % params.security.classad_proxy)
```

creation/lib/cvWCreate.py
```diff
@@ -60,7 +60,11 @@ def create_client_mapfile(mapfile_fname, my_DN, factory_DNs, schedd_DNs, collect
 
     """
     with open(mapfile_fname, "w") as fd:
-        fd.write('GSI "^{}$" {}\n'.format(re.escape(my_DN), "me"))
+        try:
+            # issue 66 transition from GSI, keep going if no DNs for proxies
+            fd.write('GSI "^{}$" {}\n'.format(re.escape(my_DN), "me"))
+        except:
```
mambelli, line 66: Try to avoid generic `except`. What is causing the problem? check if my_DN is None

mambelli (changes requested):

creation/lib/cvWParamDict.py
ddbox: implemented

creation/lib/cvWParamDict.py
ddbox: done

creation/lib/cvWCreate.py
ddbox: rewrote checking my_DN instead of using broad try:except:

mambelli (approved):

PR #187, opened 2022-07-20
Aws gce mods

ddbox:
changes to factory to send IDTOKEN to AWS in format that the VM can use.

Intended for use with cloud-vms 2.3.rc1 or later, see cloud-vms/issue/1


BrunoCoimbra (approved):
All looks good.

mambelli:
These changes seem to support only tokens. Do these need to be coordinated w/ the VM image changes? Any way to support both images (i.e. send both token and proxy if available)?

PR #190, opened 2022-07-29
fixing loop lib_fork

j00961415:


mambelli:
More complete fix in #193

PR #195, opened 2022-08-09
analyze_entries: make it python3 ready. Fixes #194

### LINKED ISSUE #194: A couple of issues  with analyze_entries and python3
It surns out that in python2 you could use urllib to open a local file (`/var/lib/gwms-factory/work-dir/monitor/rrd_Log_Completed.xml` in this case), but with python3 you need to inser `file://` to avoid this error:

```ValueError: unknown url type: '/var/lib/gwms-factory/work-dir/monitor/rrd_Log_Completed.xml'```

Also, you cannot use `.sort()` on the result of a `dict.keys()` since this is a view and not an iterator now. I was getting:

```AttributeError: 'dict_keys' object has no attribute 'sort'```

### PR DESCRIPTION
**Pull request by mmascher**

### CODE REVIEW

File: factory/tools/analyze_entries [hunk 1]
```diff
@@ -564,8 +564,7 @@ Glidein log analysis for frontend %s - %s
     ######################################################################
 
     # sort periods from least to greatest, with 24 hours at the top
-    period_list = period_data.keys()
-    period_list.sort()
+    period_list = sorted(period_data.keys())
```
mambelli, line 567: Should it be explicitly converted to list?
`period_list = sorted(list(period_data.keys()))`
or is sorted taking care of it?

mmascher: Sorted takes care of it

mambelli (approved):

PR #203, opened 2022-08-23
Adding a space after attribute name in gconfig_get to avoid partial name hits

mambelli:
Adding a space after attribute name in gconfig_get to avoid partial name hits
Added also unit tests for gconfig_get


ddbox:
fix tested and works

ddbox (approved):
changes look good

PR #204, opened 2022-08-24
Cleanout to remove the working and tmp directories also when the start dir is missing

### LINKED ISSUE #189: Missing startup_dir results in glidein_startup.sh file leak
In [glidien_cleanup](https://github.com/glideinWMS/glideinwms/blob/master/creation/web_base/glidein_startup.sh#L405-L427) (sic), cleanup gets skipped if the glidein can't `cd` into the `startup_dir`. This logic was added in a big refactor commit https://github.com/glideinWMS/glideinwms/commit/79cb6ff2360fd15ed580329cd4ddf5ec680a2cb7 but there doesn't seem to be any indication of why it was added.

An [HTCondor bug](https://opensciencegrid.atlassian.net/browse/HTCONDOR-1223) appears to result in the CE removing the `startup_dir` shortly after the job gets the SIGTERM so we end up leaking user job sandboxes.

We think that the glidein should continue with cleanup, even if the `startup_dir` doesn't exist

mambelli:
Cleanout removes the Glidein working and tmp directories also when the start dir is missing

This avoids leaking sandbox files, but status and Glidein termination info are lost because stdout and stderr are no more accessible.
This fixes #189

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Changes since the last release
 ### Changed defaults / behaviours
 
 -   Frontend configuration valid (reconfig/upgrade successful) even if some HTCSS schedds are not in DNS. Failing only if all schedds are unknown to DNS.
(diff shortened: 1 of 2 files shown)


mmascher (approved):
Looks good to me

PR #211, opened 2022-09-09
Fixed python3 check return value in case of exception

mambelli:
Fixed python3 check return value in case of exception and minor pep8 improvements


namrathaurs (approved):
Changes look fine. Ready to merge.

PR #212, opened 2022-09-13
Fixed list intersection to work w/ python2 or python3

mambelli:
Fixed list intersection to work w/ python2 or python3 (no more just python)


terranovaa (changes requested):
Correct.

terranovaa (dismissed):

PR #214, opened 2022-09-14
HTCSS DC_DAEMON_LIST set explicitly in all GWMS components except the Factory

### LINKED ISSUE #205: General GWMS config should not set DC_DAEMON_LIST
**Describe the bug**
The assumption that `DC_DAEMON_LIST` should match the entire `DAEMON_LIST` is flawed here:
https://github.com/glideinWMS/glideinwms/blob/e2f62c9b491795ab052916b95c9b0d17614e57c0/install/templates/00_gwms_general.config#L26
For example, HTCSS provides the `condor_credmon_oauth` and `condor_credmon_vault` which are HTCondor "daemons" but they are not HTCondor DaemonCore daemons (they don't take the DaemonCore command-line arguments, nor do they open a port to accept DaemonCore inter-process commands). These are often installed on access points for managing user credentials for payload jobs. The consequence of a non-DC process ending up in `DC_DAEMON_LIST` is that (1) the process will be passed unexpected arguments and either fail to start or do something else unexpected and (2) the process will not be killed properly, since the `condor_master` will try to use other forms of communication instead of sending `SIGQUIT`. This then causes further problems when condor is restarted because it gets stuck waiting for the non-DC process to go away.

**To Reproduce**
1. Install the `glideinwms-minimal-condor` package (which contains `00_gwms_general.config`), and then either of the `condor-credmon-oauth` or `condor-credmon-vault` packages.
2. Start up condor (`condor_master` or `systemctl start condor`).
3. Note that the credmon log gets written to the strangely-named file `/var/log/ocal-name`.
4. Try to restart condor (`condor_restart` or `systemctl restart condor`)
5. Note that daemons are stuck waiting for the `condor_credmon_oauth` (or `_vault`) process to stop, either by looking at the `MasterLog` or checking `condor_who -quick`. Note that the process is still running using `ps` (or other tool of choice). The restart will only complete after sending a `SIGKILL` signal to the credmon process.

**Expected behavior**
The credmon log should be written to `/var/log/CredMonOAuthLog`, and a restart should not wait more than a couple seconds for the credmon to exit.

**Screenshots**
n/a

**Desktop (please complete the following information):**

-   GlideinWMS version: n/a
-   Python version:: n/a
-   OS version: Observed on RHEL 7-based and RHEL 8-based distributions
-   HTCondor version: Tested on 9.0.16 and 9.11.0

**Additional context**
Related HTCondor tickets (somewhat obsoleted after figure out the problem was related to `DC_DAEMON_LIST`):
* https://opensciencegrid.atlassian.net/browse/HTCONDOR-1276
* https://opensciencegrid.atlassian.net/browse/HTCONDOR-1272

CC @GregThain @brianhlin @bbockelm

mambelli:
HTCSS DC_DAEMON_LIST set explicitly in all GWMS components except the Factory

This avoids including in DC_DAEMON_LIST new HTCSS daemons not using the DC library
The Factory still has `DC_DAEMON_LIST = $(DAEMON_LIST)`
We cannot rely on the HTCSS default/automatic DC_DAEMON_LIST because secondary collectors and schedds used in GWMS are not added to it.

This fixes #205


mmascher (approved):
Changes look good, I think I understand the logic even though I am not sure I fully understand the original problem. I only have a doubt. Even using SECCOLL_LIST instead of DAEMON_LIST I will be able to query secondary collector with condor_status, right?

mambelli:
> Changes look good, I think I understand the logic even though I am not sure I fully understand the original problem. I only have a doubt. Even using SECCOLL_LIST instead of DAEMON_LIST I will be able to query secondary collector with condor_status, right?

You could. Anyway, SECCOLL_LIST is just a temporary variable to add the same list to DAEMON_LIST and DC_DAEMON_LIST. You should keep using DAEMON_LIST like before to query all collectors

PR #216, opened 2022-09-14
Fixed exit code inconsistency when using tokens and made code more robust

mambelli:
Fixed exit code inconsistency when using tokens and made code more robust

Exit code consistent w/ erro_gen XML, restored check on needed condor variables and fixed x509 conditional checks, made compliant w/ new gconfig_add/get, linting improvements

### DIFF
--- a/creation/web_base/condor_startup.sh
+++ b/creation/web_base/condor_startup.sh
@@ -3,15 +3,8 @@
 # SPDX-FileCopyrightText: 2009 Fermi Research Alliance, LLC
 # SPDX-License-Identifier: Apache-2.0
 
-#
-# Project:
-#   glideinWMS
-#
-# File Version:
-#
 # Description:
-#   This script starts the condor daemons expects a config file as a parameter
-#
+#   This script starts the condor daemons. Expects a config file (glidien_config) as a parameter
 
 trap_with_arg() {
     func="$1" ; shift
@@ -45,21 +38,13 @@ ignore_signal() {
     echo "Condor startup received SIGHUP signal, ignoring..."
 }
 
-#if an IDTOKEN is available, continue.  Else exit
-exit_if_no_token(){
-    if [ !  -e "$GLIDEIN_CONDOR_TOKEN" ]; then
-        exit $1
-    fi
-    echo "condor_startup.sh" "found" "$GLIDEIN_CONDOR_TOKEN" "continuing"
-}
-
 metrics=""
 
 # put in place a reasonable default
 GLIDEIN_CPUS=1
 
 # first of all, clean up any CONDOR variable
-condor_vars=`env |awk '/^_[Cc][Oo][Nn][Dd][Oo][Rr]_/{split($1,a,"=");print a[1]}'`
+condor_vars=$(env |awk '/^_[Cc][Oo][Nn][Dd][Oo][Rr]_/{split($1,a,"=");print a[1]}')
 for v in $condor_vars; do
     unset $v
 done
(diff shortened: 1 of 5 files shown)


ddbox (changes requested):
I note that create_mapfile.sh, as submitted for review reveals a "gotcha" that should be at least documented for future retrofitters putting gconfig_* into all these glidein scripts.  It would be better to fix, I don't know if the execution environment permits it.

A shell script that uses the gconfig functions goes through a setup procedure like so:
glidein_config="$1"
add_config_line_source="$(grep '^ADD_CONFIG_LINE_SOURCE ' "$glidein_config" | cut -d ' ' -f 2-)"
. "$add_config_line_source"

The potential problem here is that glidein_config is a global_variable used by gconfig_add .

In create_mapfile.sh the same pattern  is used only with a different variable name
config_file="$1"
 add_config_line_source="$(grep '^ADD_CONFIG_LINE_SOURCE ' "$config_file" | cut -d ' ' -f 2-)"
and so on.
Things seem fine until gconfig_add is called, which causes the calling script to exit with the helpful error 
[root@fermicloud528 glide_RD05pn.save]# gconfig_add FOO BAR
Wed Sep 21 21:37:57 UTC 2022 Error: glidein_config not defined
logout
Connection to fermicloud528.fnal.gov closed.


The gconfig_add problem can be worked around by inserting a definition of the global variable 'glidein_config' like so:
config_file="$1"
glidein_config="$config_file"
add_config_line_source=.....

I don't know if this workaround introduces other problems, but there are lots of scripts that use 'config_file' instead of 'glidein_config' as a variable name, and this pattern is likely to catch retrofitters in the future

mambelli:
Thanks, I fixed the bug. The previous create_mapfile.sh code had
```
config_file="$1"
glidein_config="$1"  # TODO: why this? For some sourced file?
```
I removed the second line not understanding the need. The crash would have been there also w/ the previous add_config_line.

`add_config_line()` already required `$glidein_config` to be defined in the environment, which is not the best choice for an API. I added also a note on top of add_config_line.source, to report in the documentation.
I'm wondering if I should make a new function `gconfig_set()` where the first parameter has to be the file name. Changing the invocation of add_config_line or gconfig_add would be even more confusing. What do you think? Should we open an issue about it?

Existing scripts w/ a different config file name use only gconfig_get which takes the file name as optional second command line parameter.

creation/web_base/condor_startup.sh
```diff
@@ -216,18 +222,18 @@ set_var() {
...
-        elif [ "$var_user" == "@" ]; then
+        elif [[ "$var_user" == "@" ]]; then
             var_user=$var_condor
         fi
 
         condor_env_entry="$var_user=$var_val"
-        condor_env_entry=`echo "$condor_env_entry" | awk "{gsub(/\"/,\"\\\\\"\\\\\"\"); print}"`
-        condor_env_entry=`echo "$condor_env_entry" | awk "{gsub(/'/,\"''\"); print}"`
-        if [ -z "$job_env" ]; then
+        condor_env_entry=$(echo "$condor_env_entry" | awk "{gsub(/\"/,\"\\\\\"\\\\\"\"); print}")
```
ddbox, line 234: lines 234 and 235  fail after converting from backtick to $() notation.

[root@fermicloud528 glide_pvw10X]# condor_env_entry=`echo "$condor_env_entry" | awk "{gsub(/\"/,\"\\\\\"\\\\\"\"); print}"`
[root@fermicloud528 glide_pvw10X]# echo $condor_env_entry
""an oddly quoted string
[root@fermicloud528 glide_pvw10X]# condor_env_entry='"an oddly quoted string'
[root@fermicloud528 glide_pvw10X]# condor_env_entry=$(echo "$condor_env_entry" | awk "{gsub(/\"/,\"\\\\\"\\\\\"\"); print}")
awk: cmd. line:1: {gsub(/"/,"\\"\\""); print}
awk: cmd. line:1:               ^ backslash not last character on line
awk: cmd. line:1: {gsub(/"/,"\\"\\""); print}
awk: cmd. line:1:               ^ syntax error

mambelli: fixed using `condor_env_entry=$(echo "$condor_env_entry" | awk '{gsub(/"/,"\"\""); print}')` :
```
bash-3.2$ condor_env_entry='"an oddly quoted string'
bash-3.2$ condor_env_entry=$(echo "$condor_env_entry" | awk '{gsub(/"/,"\"\""); print}')
bash-3.2$ echo $condor_env_entry
""an oddly quoted string
```

mambelli: I double checked also the single quote one that was OK

ddbox (approved):

PR #221, opened 2022-09-19
Issue 182

dbox-fnal:
send both proxy and idtoken to gce and ec2 images.  AMI to test against is ami-0e07f2c24c840cfcd

### DIFF
--- a/creation/lib/cvWParamDict.py
+++ b/creation/lib/cvWParamDict.py
@@ -1364,7 +1364,7 @@ def populate_group_security(client_security, params, sub_params, group_name):
 
                 if (pel["pool_idx_len"] is None) and (pel["pool_idx_list"] is None):
                     # only one
-                    dn = x509Support.extract_DN(proxy_fname)
+                    dn = x509Support.extract_DN(proxy_fname, exit_if_bad=False)
                     # don't worry about conflict... there is nothing wrong if the DN is listed twice
                     pilot_dns.append(dn)
                 else:
--- a/creation/lib/cvWParamDict.py
+++ b/creation/lib/cvWParamDict.py
@@ -1363,10 +1363,13 @@ def populate_group_security(client_security, params, sub_params, group_name):
                     proxy_fname = pel["pilotabsfname"]
 
                 if (pel["pool_idx_len"] is None) and (pel["pool_idx_list"] is None):
-                    # only one
-                    dn = x509Support.extract_DN(proxy_fname, exit_if_bad=False)
-                    # don't worry about conflict... there is nothing wrong if the DN is listed twice
-                    pilot_dns.append(dn)
+                    try:
+                        # only one
+                        dn = x509Support.extract_DN(proxy_fname)
+                        # don't worry about conflict... there is nothing wrong if the DN is listed twice
+                        pilot_dns.append(dn)
+                    except SystemExit:
+                        print("...Failed to extract DN from %s, but continuing" % proxy_fname)
                 else:
                     # pool
                     pool_idx_list_expanded_strings = get_pool_list(pel)
--- a/factory/glideFactoryLib.py
+++ b/factory/glideFactoryLib.py
@@ -1940,6 +1940,8 @@ def get_submit_environment(
             # Hence, use " for now.
             exe_env.append("GLIDEIN_ARGUMENTS=%s" % glidein_arguments)
         elif grid_type in ("ec2", "gce"):
+            # do not uncomment these in production, it exposes secrets that
+            # should not be logged
             # log.debug("params: %s" % str(params))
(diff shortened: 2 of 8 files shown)


factory/glideFactoryLib.py
```diff
@@ -1940,9 +1940,9 @@ def get_submit_environment(
             # Hence, use " for now.
             exe_env.append("GLIDEIN_ARGUMENTS=%s" % glidein_arguments)
         elif grid_type in ("ec2", "gce"):
-            log.debug("params: %s" % str(params))
-            log.debug("submit_credentials.security_credentials: %s" % str(submit_credentials.security_credentials))
-            log.debug("submit_credentials.identity_credentials: %s" % str(submit_credentials.identity_credentials))
+            # log.debug("params: %s" % str(params))
```
mambelli, line 1943: Add a comment line like:
```
# Sensitive information. Uncomment this block only on private development machines
```

factory/glideFactoryLib.py
```diff
@@ -1982,6 +1982,7 @@ def get_submit_environment(
                 ini_template = """[glidein_startup]
 args = %s
 idtoken_file_name = credential_frontend.idtoken
+proxy_file_name = pilot_proxy
 webbase= %s
 
 [vm_properties]
```
mambelli, line 1988: Maybe change to use string interpolation:
```
                ini_template = f"""[glidein_startup]
args = {glidein_arguments}
idtoken_file_name = credential_frontend.idtoken
proxy_file_name = pilot_proxy
webbase= {web_url}
...
```

factory/glideFactoryLib.py
```diff
@@ -2008,6 +2009,15 @@ def get_submit_environment(
...
                 exe_env.append("GLIDEIN_PROXY_FNAME=%s" % full_path_to_proxy)
 
+                # get the IDTOKEN
+                token_data = ""
+                with open(submit_credentials.identity_credentials["frontend_condortoken"]) as fc:
+                    tok = fc.readlines()
+                    for tln in tok:
+                        token_data = f"{token_data}{tln}"
+
+                exe_env.append("IDTOKEN_DATA=%s" % token_data)
```
mambelli, line 2019: Adding the token content to the environment there could be the risk of someone looking at it. The Factory should be a secure server, still, the file would be safer. Would be difficult to do something similar to the proxy, using the file name?

lib/x509Support.py
```diff
@@ -8,7 +8,7 @@
 from . import defaults
 
 
-def extract_DN(fname):
+def extract_DN(fname, exit_if_bad=True):
```
mambelli, line 11: This library is called in many places. 
If the file is missing or there is a different content probably it should not be called instead of ignoring the error.
I think the problem was to have an sys.exit inside a library function.
I'd prefer to:
1. leave `def extract_DN(fname):`
2. change original `sys.exit(3)` to a `raise` (re-raising ValueError)
3. catch ValueError in the caller
Would be nice to add a message also to the log file, but I'm not sure the logger is available

creation/lib/cvWParamDict.py
```diff
@@ -1364,7 +1364,7 @@ def populate_group_security(client_security, params, sub_params, group_name):
 
                 if (pel["pool_idx_len"] is None) and (pel["pool_idx_list"] is None):
                     # only one
-                    dn = x509Support.extract_DN(proxy_fname)
+                    dn = x509Support.extract_DN(proxy_fname, exit_if_bad=False)
```
mambelli, line 1367: Put this in a try except or protect it with an if (see x509Support below):
```
                    try:
                        dn = x509Support.extract_DN(proxy_fname)
                    except ValueError:
                        pass
OR
                    if os.path.exists(proxy_fname):
                        dn = x509Support.extract_DN(proxy_fname)
                    else:
                        dn = ""
```

mambelli (changes requested):

creation/lib/cvWParamDict.py
dbox-fnal: put in a try/except as suggested

factory/glideFactoryLib.py
dbox-fnal: removed from exe environment visibility.  Required a re-working of credential compression to check if token, proxy, or both provided

factory/glideFactoryLib.py
dbox-fnal: appropriate comments added

factory/glideFactoryLib.py
dbox-fnal: reasonable suggestion but I did not implement, it works as is

mambelli (approved):

PR #223, opened 2022-10-10
Adding the possibility to set TRUST_DOMAIN via the attr TRUST_DOMAIN

mambelli:
Adding the possibility to set TRUST_DOMAIN via the attr TRUST_DOMAIN

By default TRUST_DOMAIN is GLIDEIN_Collector and if that is empty, CCB_ADDRESS.
Now if TRUST_DOMAIN is set it will be checked and used first.


namrathaurs (approved):
Changes look fine according to the description of the PR. Ready for merging.

mambelli:
Obsoleted from PR #226

PR #233, opened 2022-11-01
fix broken uses of PATH (SOFTWARE-5340)

edquist:
"unset PATH" results in regular shell commands like grep becoming broken, and everything falls apart.

Also, there is one spot where it appears LD_LIBRARY_PATH was intended but PATH was used.

FIX! FIX! FIX!


mambelli (approved):

mambelli:
Thanks, Carl!

PR #235, opened 2022-11-13
Updated singularity_lib.sh to use gconfig_ functions and bug fixes and hardenings

mambelli:
Updated singularity_lib.sh to use gconfig_ functions and bug fixes and hardenings

Hardenings
- checking also singularity exit code besides the output in singularity_test_exec
- inspecting singularity stderr and raising a warning if there are "FATAL" errors and the exit code is 0

Bugs fixed:
- return correctly default value in gwms_from_config when it should. Added also unit test
- fixed bug in get_prop_str: "undefined" returns the default value
- added tac to the mac compatibility commands for BATS

NOTE: using gconfig_get returns only one line (before was returning all the matching lines)


namrathaurs (dismissed):
Changes look good and are ready for merging!

namrathaurs (approved):
Revised changes look fine and can be merged.

PR #237, opened 2022-11-21
Added python utilities to read and write glidien_config

mambelli:
Added python utilities to read and write glidien_config. Use `import gconfig` when using `gwms-python`.
`gconfig.py` is in the glidein `lib/python` directory (and PYTHONPATH) since is compatible w/ both python2 and python3.
To use gwms-python and gconfig the scripts must be in the after-entry priority group.

Fixed also a couple of typos in comments and added changelog entries

TODO:
- there are unit tests for get. More tests should be added

This provides functionalities similar to https://gitlab.cern.ch/CMSSI/CMSglideinWMSValidation/-/blob/production/export_siteconf_info.py
The CMS implementation has some differences from the shell implementation and bugs:
- overwrites all keys starting with the key string (app -> application)
- removes trailing spaces


mmascher (approved):
LGTM

PR #241, opened 2022-11-23
Added autodetect for aarch64 (ARM) architecture in condor platforms

mambelli:
Added autodetect for aarch64 (ARM) architecture in condor platforms

Manual selection works also for unknown architectures


mmascher (approved):
Looks good to me

PR #244, opened 2022-11-29
Adding support for glidein to use cvmfsexec in mode 3 when possible 

namrathaurs:
While this PR primarily encompasses changes pertaining to enabling support for using mode 3 of cvmfsexec whenever possible, it also contains an updated way to check for script duplication.


namrathaurs:
Closing this PR since the upstream branch was outdated at the time of creating this draft PR, which resulted in a huge number of changes/files to be reviewed which are merged with the upstream master.

PR #251, opened 2022-12-02
Fixed tags not to have " #" which starts a comment in yaml

mambelli:
Fixed tags not to have " #" which starts a comment in YAML also if not at the beginning of a line. 
Inside a quoted string or just "#" without a space before is OK.
Improved also process_tags code quality


namrathaurs (approved):
Looks good and is ready for merging!

PR #252, opened 2022-12-06
Fix for failing GWMS rpms build jobs

namrathaurs:
When contributors push commits to their forked GWMS repository, one of the Github workflows that runs a check for building GWMS RPMs was failing. This PR provides the fix for the same.

Upon inspecting the log file corresponding to the rpm build workflow, the following error was reported:
```
release.py: error: no such option: --rpm-version
```

The usage documentation for `release.py` does not include/mention any option named `rpm-version`. The `release.py` script when invoked as part of the Github action, among other options, included an option named `rpm-version` even though the script did not support such an option. Removing the option from the invocation line results in the build workflow to complete successfully without any further errors.


namrathaurs:
PR is now ready for review. The failing pyunittest `test_lib_symCrypto.py` can be safely ignored (as per Marco's input) as there is already a safety check included in the code for strings starting with `b'` to be flagged as error)

mambelli (approved):

PR #260, opened 2023-01-24
Disabling TRACK and TRACE in the web server configuration

mambelli:
Disabling TRACK and TRACE in the web server configuration as recommended for production.
They are not used by GWMS and may be a vehicle to attack the web server:
https://www.virtuesecurity.com/kb/web-server-trace-enabled/

### DIFF
--- a/build/packaging/rpm/gwms-factory.conf.httpd
+++ b/build/packaging/rpm/gwms-factory.conf.httpd
@@ -18,6 +18,23 @@ Alias /factory /var/lib/gwms-factory/web-area
 </Directory>
 
 
+# TRACE and TRACK are not used in GlideinWMS and not recommended in production
+# Add the following line in httpd.conf to disable trace on the whole server
+# TraceEnable Off
+#
+# The lines below will disable unused methods using mod_rewrite
+# Use the following RewriteCondition to just disable trace/track and
+# allow more method beside get/post/head
+#     RewriteCondition %{REQUEST_METHOD} ^(TRACE|TRACK)
+# or comment completely if you need also track and trace or
+# mod_rewrite uses too much CPU
+<IfModule mod_rewrite.c>
+    RewriteEngine on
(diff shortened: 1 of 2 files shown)


build/packaging/rpm/gwms-factory.conf.httpd
```diff
@@ -18,6 +18,22 @@ Alias /factory /var/lib/gwms-factory/web-area
 </Directory>
 
 
+# TRACE and TRACK are not used in GlideinWMS and not recommended in production
+# Add the following line in httpd.conf to disable trace on the whole server
+# TraceEnable Off
+# This will disable unused methods using mod_rewrite
```
namrathaurs, line 24: Is this comment for the `TraceEnable` option or the `RewriteCondition`? Since it's in between the two, personally it seems a bit confusing as to which this comment is applicable

build/packaging/rpm/gwms-frontend.conf.httpd
```diff
@@ -18,6 +18,23 @@ Alias /vofrontend /var/lib/gwms-frontend/web-area
     Options -Indexes
 </Directory>
 
+
+# TRACE and TRACK are not used in GlideinWMS and not recommended in production
+# Add the following line in httpd.conf to disable trace on the whole server
+# TraceEnable Off
+# This will disable unused methods using mod_rewrite
```
namrathaurs, line 25: same comment as previous

build/packaging/rpm/gwms-factory.conf.httpd
mambelli: Text improved to make more clear that refers to the following lines

build/packaging/rpm/gwms-frontend.conf.httpd
mambelli: same

namrathaurs (approved):
changes can be merged

PR #265, opened 2023-01-26
Make sure to call add_identity_credential for scitoken. Fixes #264

### LINKED ISSUE #264: project_id does not work with scitokens
**Describe the bug**
Colby reported the factory was not sending pilots to an entry that was supposed to use `scitoken+projectid` as `auth_method`. Found this exception in the entry logfile in the factory:

```Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/glideinwms/factory/glideFactoryLib.py", line 2018, in get_submit_environment
    glidein_rsl = "{}(project={})".format(glidein_rsl, submit_credentials.identity_credentials["ProjectId"])
KeyError: 'ProjectId'
```

**To Reproduce**
Setup an entry with `scitoken+projectid`, use that auth method in the frontend and try to submit.

**Expected behavior**
Pilots should be submitted with no exception

-   Priority: critical
-   Stakeholder: FactoryOps
-   Component: factory

mmascher:


factory/glideFactoryEntry.py
```diff
@@ -1300,6 +1300,19 @@ def unit_work_v3(
                 % (scitoken_file, str(submit_credentials.identity_credentials))
             )
 
+    if ("grid_proxy" in auth_method) or ("scitoken" in auth_method):
+        # Check if project id is required
```
mambelli, line 1304: Should this be removed? To add the project_id if in auth methods no matter which other auth method is there?

mmascher: Yes, I think it makes sense. If anything we should check the `grid_type` since here https://github.com/glideinWMS/glideinwms/blob/366c9481b81392f5cfb1168257d64cc7516dd928/factory/glideFactoryLib.py#L2043 we add the `project_id` to the rsl and the environment only if the grid type is NOT `ec2`, `gce` and `batch`. But I think it does not really matter since at reconfig time we only use the project id for `condor`: https://github.com/glideinWMS/glideinwms/blob/25e66daac4d0a2a789f1c03617c6f5758dd2c924/creation/lib/cgWCreate.py#L315

So, I would say let's always add it here since it does not hurt. This way we are sure that if `"project_id" in auth_method` then `ProjectId in submit_credential` and avoid the issue of the ticket.

I will make the change, let me know if there is anything else.

mambelli (approved):

PR #269, opened 2023-03-06
Add missing error code for BEST_FIT in OSG_autoconf

mmascher:


mmascher:
This comes from an incomplete commit long time ago. The factory was patched in production.

mambelli (approved):

PR #271, opened 2023-03-06
Improve the generation of JDL file for ARC CE REST interface

mmascher:


creation/lib/cgWCreate.py
```diff
@@ -171,8 +171,9 @@ def populate(self, exe_fname, entry_name, conf, entry):
         enc_input_files.append("$ENV(IDTOKENS_FILE:)")
         self.add_environment("IDTOKENS_FILE=$ENV(IDTOKENS_FILE:)")
 
-        if gridtype not in ["ec2", "gce"]:
+        if gridtype not in ["ec2", "gce"] and not (gridtype=="arc" and auth_method=="grid_proxy"):
             self.add("+SciTokensFile", '"$ENV(SCITOKENS_FILE:)"')
+        self.add("+x509UserProxy", '"$ENV(X509_USER_PROXY:)"')
```
mambelli, line 176: Adding this for all.
Should it check if X509_USER_PROXY is defined?

mambelli (changes requested):

creation/lib/cgWCreate.py
mmascher: Yes, adding this for all because I am not sure if the nmew condor behaviour (ignore the X509_USER_PROXY variable) is only for arc or everything. In any case for condor-ce the usage of GSI or TOKEN is negotiated between condor_submit running on the factory and the condor collector daemon running on the CE. We still need to propagate the proxy for the time being (even with condor 10)

About checking if X509_USER_PROXY is defined. This code is adding a line in the JDL file at reconfig time. The check should be done by the factory daemons when the real condor_submit is done. It is probably better to not add the 509UserProxy if the variable is not defined. Not sure how to do that however.. Any sugestion?

mambelli: I'm worried that this may cause problems for people not using proxies.
Could you check a couple of things (ask the condor team if not sure)?
1. The [condor documentation](https://htcondor.readthedocs.io/en/latest/man-pages/condor_submit.html?highlight=x509UserProxy#submit-description-file-commands) mentions `x509userproxy`, not `+x509UserProxy` (user variable)
2. The same document says it uses `$ENV(X509_USER_PROXY)` if that attribute is not set in the submit file, so do you need it?
3. What happens if the `$X509_USER_PROXY` is empty (you'd have `+x509UserProxy = ` )? Supposing that there are all the tokens to submit and for the Glidein to call back, Is this just ignored or is it there a submission error?
4. What heppens if the referred path does not exist (always all needed tokens are there)?
5. Ad what happens if the pointed proxy is expired (again all needed tokens are there)?

I'd be OK in 5 causing an error, but I'd like 3 and 4 not to fail submission (if things can continue w/ the provided tokens).
If you are sure of the syntax (1,2) and that there are no problems please confirm here and I'll approve the merge. Thank you

mmascher: 1. It seems there is a bug in condor when you use `x509userproxy` without the plus and grid resource `arc`:

```
[0737] root@gfactory-itb-1 /tmp# echo $X509_USER_PROXY 
/var/lib/gwms-factory/client-proxies/user_fecmsglobalitb/glidein_gfactory_instance/credential_CMSG-ITB_gWMSFrontend-v1_0.main_7ZXLEJW7
[0738] root@gfactory-itb-1 /tmp# grep x509 job.arc 
x509UserProxy = "$ENV(X509_USER_PROXY:)"
[0738] root@gfactory-itb-1 /tmp# sudo -Eu gfactory condor_submit job.arc
Submitting job(s)WARNING: GSI authentication is enabled by your security configuration! GSI is no longer supported.
For details, see https://htcondor.org/news/plan-to-replace-gst-in-htcss/
ERROR: unable to read proxy file
[0738] root@gfactory-itb-1 /tmp# vim job.arc 
[0738] root@gfactory-itb-1 /tmp# grep x509 job.arc 
+x509UserProxy = "$ENV(X509_USER_PROXY:)"
[0738] root@gfactory-itb-1 /tmp# sudo -Eu gfactory condor_submit job.arc
Submitting job(s)WARNING: GSI authentication is enabled by your security configuration! GSI is no longer supported.
For details, see https://htcondor.org/news/plan-to-replace-gst-in-htcss/
.
1 job(s) submitted to cluster 313.
[0738] root@gfactory-itb-1 /tmp# condor_q -l 313 | grep -i x509
WARNING: GSI authentication is enabled by your security configuration! GSI is no longer supported.
For details, see https://htcondor.org/news/plan-to-replace-gst-in-htcss/
x509UserProxy = "/var/lib/gwms-factory/client-proxies/user_fecmsglobalitb/glidein_gfactory_instance/credential_CMSG-ITB_gWMSFrontend-v1_0.main_7ZXLEJW7"
x509UserProxyExpiration = 1678950007
x509userproxysubject = "/DC=ch/DC=cern/OU=computers/CN=cmspilot04/vocms0802.cern.ch"
[0739] root@gfactory-itb-1 /tmp#
```

I even used strace and verified condor was trying to open the wrong file:

```
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f8de110b000
read(5, "#!/bin/bash\n\necho \"ciao\"\nsleep 10s\nhostname\n", 4096) = 44
close(5)                                = 0
munmap(0x7f8de110b000, 4096)            = 0
open("/tmp/\"/var/lib/gwms-factory/client-proxies/user_fecmsglobalitb/glidein_gfactory_instance/credential_CMSG-ITB_gWMSFrontend-v1_0.main_7ZXLEJW7\"", O_RDONLY) = -1 ENOENT (No such file or directory)
write(2, "ERROR: unable to read proxy file\n", 33ERROR: unable to read proxy file
) = 33
exit_group(1) 
```

I have have not gotten confirmation from Jaime about this bug though, to be honest it slipped my mind, I just put the `+` and forgot... I'll send him an email with you in CC.

For grid resource `condor` the version without the `+` works fine.

2.  I think the condor documentation is out of date. From my tests if I do not explicitly put x509userproxy in the JDL it does not work on either grid resource types. I remember asking Jaime about this at one of the past HTCodnor-CMS meetings, he confirmed it.

About 3/4/5. I actually just noticed that for HTCondor we were already setting x509_user_proxy: https://github.com/glideinWMS/glideinwms/blob/master/creation/lib/cgWCreate.py#L346

The difference between `x509_user_proxy` and `+x509_user_proxy` is that the first one is a macro and does some checking. Therefore if the file does not exist or it is expired the submission fails. With the `+` the submission succeds (but the job will likely go held anyway at a later stage..)

mmascher: FYI: given the above I changed the commit to only add the proxy for arc grid type

mmascher: Final version after discussion with @mambelli . Turned out the `x509UserProxy` macro (no `+`) can be used if you do not put `"` around the filename.

mambelli (approved):

Issue #262, opened 2023-01-25
Factory Knob to Control Rebuilding of cvmfsexec Distributions


namrathaurs:
**Is your feature request related to a problem? Please describe.**
Reconfiguring factories is taking too long, anywhere between 30-60 minutes, to complete the rebuilding of cvmfsexec distributions. This was reported by Jeff Dost during January 2023's stakeholders' meeting.

**Describe the solution you'd like**
Add a formal knob/trigger in the factory that would provide more control, i.e. disable/enable as needed, over the rebuilding of cvmfsexec distributions. This is useful for scenarios when the cvmfs is to be provisioned on demand for a specific entry (as opposed to all the entries defined in the configuration file of the factory).

**Describe alternatives you've considered**
Remove the script, specified as a pre-reconfig hook, that builds these cvmfsexec distributions.

**Info (please complete the following information):**

-   Priority: High
-   Stakeholder: FactoryOps
-   Component: Factory

**Additional context**
The latest version of the reconfig hook rebuilds the cvmfsexec distributions only when a new version of cvmfsexec package is available OR when factory reconfig/upgrade is initiated the very first time. Otherwise, the rebuilding of distributions is ignored and existing versions of the distributions, that are available on the factory, are shipped to the entries/CEs by the glidein(s).

mambelli:
As a workaround, the build may be disabled by moving `create_cvmfsexec_distros.sh` out of the `/etc/gwms-factory/hooks.reconfig.pre` directory

PR #272, opened 2023-03-07
Factory Knob for Controlling Cvmfsexec Distributions

namrathaurs:
Addresses #262 to add improved support for controlling the (re)building of cvmfsexec distributions using a new knob in factory configuration file (hyperlinked issue for more details).

### DIFF
--- a/creation/create_cvmfsexec_distros.sh
+++ b/creation/create_cvmfsexec_distros.sh
@@ -10,116 +10,130 @@ usage() {
 	echo "the source to download the latest cvmfs configuration and repositories."
 }
 
-start=`date +%s`
+build_cvmfsexec_distros() {
+	start=$(date +%s)
 
-CVMFS_SOURCES=osg:egi:default
-# rhel6-x86_64 is not included; currently not supported due to EOL
-# egi for rhel8-x86_64 results in an error - egi does not yet have a centos8 build (as confirmed with Dave)
-# TODO: verify the logic when egi provides a centos8 build
-SUPPORTED_TYPES=rhel7-x86_64:rhel8-x86_64:suse15-x86_64
-factory_config_file="/etc/gwms-factory/glideinWMS.xml"
+	# rhel6-x86_64 is not included; currently not supported due to EOL
+	# egi for rhel8-x86_64 results in an error - egi does not yet have a centos8 build (as confirmed with Dave)
+	# TODO: verify the logic when egi provides a centos8 build
 
-# first, checking in the work-dir location for the current version of cvmfsexec
-work_dir=$(grep -m 1 "submit" "$factory_config_file" | sed 's/.*base_dir="\([^"]*\)".*/\1/')
-# protect aginst non-existence of cvmfsexec directory; fresh install of GWMS with first run of factory upgrade
-if [[ -d "$work_dir/cvmfsexec" && -d "$work_dir/cvmfsexec/tarballs" ]]; then
-    cvmfsexec_tarballs=$work_dir/cvmfsexec/tarballs
-    if [[ -f "$cvmfsexec_tarballs/.cvmfsexec_version" ]]; then
-        curr_ver=$(cat "$cvmfsexec_tarballs"/.cvmfsexec_version)
-        echo "Current version found: $curr_ver"
-    fi
+	factory_config_file="/etc/gwms-factory/glideinWMS.xml"
+	# first, checking in the work-dir location for the current version of cvmfsexec
+	work_dir=$(grep -m 1 "submit" "$factory_config_file" | sed 's/.*base_dir="\([^"]*\)".*/\1/')
+	# protect aginst non-existence of cvmfsexec directory; fresh install of GWMS with first run of factory upgrade
+	if [[ -d "$work_dir/cvmfsexec" && -d "$work_dir/cvmfsexec/tarballs" ]]; then
+		cvmfsexec_tarballs=$work_dir/cvmfsexec/tarballs
+		if [[ -f "$cvmfsexec_tarballs/.cvmfsexec_version" ]]; then
+			curr_ver=$(cat "$cvmfsexec_tarballs"/.cvmfsexec_version)
+			echo "Current version found: $curr_ver"
+		fi
+	else
+		# if the cvmfsexec directory does not exist, create one
+		# also, create a directory named tarballs under cvmfsexec directory
+		cvmfsexec_tarballs="$work_dir/cvmfsexec/tarballs"
+		# check if tarballs directory exists; if not, create one; else proceed as usual
+		if mkdir -p "$cvmfsexec_tarballs"; then
+			chmod 755 "$cvmfsexec_tarballs"
+		else
+			# if the directory creation fails, print a message and exit from the script
+			echo "Unable to create directory $cvmfsexec_tarballs"
+			exit 1
+		fi
+	fi
+
+	# otherwise, .cvmfsexec_version file does not exist from a previous upgrade or it's a first-time factory upgrade
+	cvmfsexec_temp="$work_dir/cvmfsexec/cvmfsexec.tmp"
+	# check if the temp directory for cvmfsexec exists
+	if mkdir -p "$cvmfsexec_temp"; then
+		# cvmfsexec temp directory does not exist (create one) or exists (proceed to reuse)
+		chmod 755 "$cvmfsexec_temp"
+	fi
+
+	cvmfsexec_latest="$cvmfsexec_temp"/latest
+	CVMFSEXEC_REPO="https://www.github.com/cvmfs/cvmfsexec.git"
+	git clone $CVMFSEXEC_REPO "$cvmfsexec_latest" &> /dev/null
+	latest_ver=$("$cvmfsexec_latest"/cvmfsexec -v)
+	if [[ "$curr_ver" == "$latest_ver" ]]; then
+		# if current version and latest version are the same
+		echo "Current version and latest version of cvmfsexec are identical!"
+		if [[ -f "$cvmfsexec_tarballs/.cvmfsexec_version" ]]; then
+			echo "Using (existing) cvmfsexec version $(cat "$cvmfsexec_tarballs"/.cvmfsexec_version)"
+		fi
+		echo "Skipping the building of cvmfsexec distribution tarballs..."
+		rm -rf "$cvmfsexec_latest"
+		exit 0
+	else
+		# if current version and latest version are different
+		if [[ -z "$curr_ver" ]]; then
+			# $curr_ver is empty; first time run of factory upgrade
+			# no version info stored in work-dir/cvmfsexec/tarballs
+			echo "Building cvmfsexec distribution(s)..."
+		else
+			# $curr_ver is not empty; subsequent run of factory upgrade (and not the first time)
+			echo "Found newer version of cvmfsexec..."
+			echo "Rebuilding cvmfsexec distribution(s) using the latest version..."
+		fi
+
+		# build the distributions for cvmfsexec based on the source, os and platform combination
+		cvmfsexec_distros="$cvmfsexec_temp"/distros
+		if [[ ! -d "$cvmfsexec_distros" ]]; then
+			mkdir -p "$cvmfsexec_distros"
+		fi
+
+		cvmfs_configurations=($(echo $1 | tr "," "\n"))
+		supported_machine_types=($(echo $2 | tr "," "\n"))
+
+		for cvmfs_src in "${cvmfs_configurations[@]}"
+		do
+			for mach_type in "${supported_machine_types[@]}"
+			do
+				echo -n "Making $cvmfs_src distribution for $mach_type machine..."
+				os=$(echo $mach_type | awk -F'-' '{print $1}')
+				arch=$(echo $mach_type | awk -F'-' '{print $2}')
+				if "$cvmfsexec_latest"/makedist -m $mach_type $cvmfs_src &> /dev/null ; then
+					"$cvmfsexec_latest"/makedist -o "$cvmfsexec_distros"/cvmfsexec-${cvmfs_src}-${os}-${arch} &> /dev/null
+					if [[ -e "$cvmfsexec_distros"/cvmfsexec-${cvmfs_src}-${os}-${arch} ]]; then
+						echo " Success"
+						tar -cvzf "$cvmfsexec_tarballs"/cvmfsexec_${cvmfs_src}_${os}_${arch}.tar.gz -C "$cvmfsexec_distros" cvmfsexec-${cvmfs_src}-${os}-${arch} &> /dev/null
+					fi
+				else
+					echo " Failed! REASON: $cvmfs_src may not yet have a $mach_type build."
+				fi
+
+				# delete the dist directory within cvmfsexec to download the cvmfs configuration
+				# and repositories for another machine type
+				rm -rf "$cvmfsexec_latest"/dist
+			done
+		done
+
+		# remove the distros and latest folder under cvmfsexec.tmp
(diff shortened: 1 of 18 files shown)


mambelli (changes requested):
Nothing really incorrect, more suggestions

creation/create_cvmfsexec_distros.sh
```diff
@@ -3,123 +3,166 @@
...
-    if [[ -z "$curr_ver" ]]; then
-        # $curr_ver is empty; first time run of factory upgrade
-        # no version info stored in work-dir/cvmfsexec/tarballs
-        echo "Building cvmfsexec distributions..."
-    else
-        # $curr_ver is not empty; subsequent run of factory upgrade (and not the first time)
-        echo "Found newer version of cvmfsexec..."
-        echo "Rebuilding cvmfsexec distributions using the latest version..."
-    fi
+check_directory_exists() {
```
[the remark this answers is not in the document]
namrathaurs: agreed; the name has been updated to `ensure_directory_exists`

creation/create_cvmfsexec_distros.sh
```diff
@@ -3,123 +3,166 @@
...
+	# TODO: store/update version information in the $cvmfsexec_tarballs location for future reconfig/upgrade
+	if [[ "$successful_builds" -gt 0 ]]; then
+		# update only if there was at least one successful build of cvmfsexec
+		echo "$latest_ver" > "$cvmfsexec_tarballs"/.cvmfsexec_version
+	fi
+
+	end=$(date +%s)
+
+	runtime=$((end-start))
+	echo "Took $runtime seconds to create $successful_builds cvmfsexec distribution(s)"
```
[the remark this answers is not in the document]
namrathaurs: updated as suggested

creation/create_cvmfsexec_distros.sh
```diff
@@ -3,123 +3,166 @@
...
+	echo ""
+	echo "DIR: full, absolute path to the factory work directory"
+	echo ""
+	echo "SOURCES_LIST: specifies the source(s) to download the latest cvmfs "
+	echo "configuration and repositories from. Must be at least one value or "
+	echo "a comma-separated list of values from the options {osg|egi|default}."
+	echo ""
+	echo "PLATFORMS_LIST: indicates the machine types (platform- and architecture-based) "
+	echo "for which distributions is to be built. Must be at least one value or "
+	echo "a comma-separated list of values from the options {rhel7-x86_64|rhel8-x86_64|suse15-x86_64}."
```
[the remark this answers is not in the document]
namrathaurs: Conceptually, it can be empty; in which case, PLATFORMS_LIST will be all available platforms (default). I wrote the usage documentation, from the perspective where it will be executed as a standalone script. I remember you had mentioned that I should also consider the script being run by itself and not always being invoked from a factory reconfig standpoint. I shall update the usage documentation to reflect what I described above and change the main part of the script accordingly.

creation/create_cvmfsexec_distros.sh
```diff
@@ -3,123 +3,166 @@
...
+		work_dir="$2"
+		shift 2
+	else
+		# if --work-dir is not passed, assume default work-dir (RPM install)
+    	work_dir="$DEFAULT_WORK_DIR"
+	fi
+	[[ $# -eq 0 || $# -eq 1 ]] && usage || true
+	configurations=$1
+	machine_types=$2
+fi
```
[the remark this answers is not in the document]
namrathaurs: Talking from purely a self-invocation of the script, if only the option is used, i.e. `$0 --work-dir DIR`, it is an illegal invocation because both `SOURCES_LIST` and `PLATFORMS_LIST` are required arguments (as can be seen in the usage documentation). Even though you did mention that I need to also keep in mind that the script should be functional outside of the usual reconfig/upgrade process, I think I ended up assuming that the script is rarely going to be invoked on its own and would mostly be invoked as part of factory reconfig/upgrade.

In our previous discussion, we did decide that configurations will be required and machine types will be optional. So I agree that it is better to accept the option-only case as OK. I'll modify this part based on your suggestion. 

One question by the way. Is it possible that when an option is passed, there are chances that the value corresponding to the option might not be passed. Asking this to make sure if this is a possible scenario when a shell script is invoked. I've handled this scenario in my implementation but wanted to know if this is really needed/necessary.

creation/lib/cgWParamDict.py
```diff
@@ -318,50 +319,76 @@ def populate(self, other=None):
...
-        if len(distros) == 0:
-            print("Distributions for cvmfsexec not found... Skipping tarball creation.")
+        # Check if cvmfsexec distributions need to be built/rebuilt
+        populate_cvmfsexec_build_config(self.dicts["build_cvmfsexec"], self.conf)
+        cfgs = self.dicts["build_cvmfsexec"]["sources"]  # returns string type
+        mtypes = self.dicts["build_cvmfsexec"]["platforms"]  # returns string type
+        # validation checks already would have been done at the time of reading from the factory xml file; so variables cfgs and mtypes cannot be empty by this point in the execution
+        # framing the arguments to the subprocess wrapper as a string
+        # executing the cvmfsexec distribution building script can be done without explicitly specifying the location of the script since it is in the PATH variable (directory is /usr/bin/ which is set as the standard for the RPM installation)
+        if not (cfgs and mtypes):
```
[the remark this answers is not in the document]
namrathaurs: It makes sense to run the script all the time and not invoke a subprocess if there is nothing to do. Have updated the same accordingly.

creation/create_cvmfsexec_distros.sh
```diff
@@ -3,123 +3,185 @@
...
+			usage
+		else
+			# if false, SOURCES_LIST is empty, show usage and exit
+			configurations=$1; machine_types=$DEFAULT_MACHINE_TYPES
+		fi
+	else
+		echo "Invalid number of arguments passed!"
+		exit 1
+	fi
+fi
```
mambelli, line 184: Your script is actually accepting 0 arguments (considering a legal noop).
If the first argument is required you could use something like:
```shell
 if [[ $# -eq 0 || $# -gt 2 ]]; then
	echo "ERROR: Invalid number of arguments passed!"
    usage  # maybe remove the exit from usage if you'd like to have both exit 1 or 0 after it
	exit 1
fi
```
If you keep the noop (I lean towards this since it makes it easier to use):
```shell
 if [[ $# -eq 0 ]]; then
	echo "No sources specified. Building/Rebuilding of cvmfsexec distributions disabled!"
	exit 0
elif [[ $# -gt 2 ]]; then
	echo "ERROR: Invalid number of arguments passed!"
    usage  # maybe remove the exit from usage if you'd like to have both exit 1 or 0 after it
	exit 1
fi
```
After either one you are assured to have 1 or 2 arguments.
```shell
configurations=$1
machine_types=$(2:-$DEFAULT_MACHINE_TYPES}
```
If you control `$1` do it all the time. You can choose a positive or negative check. Or you can let later in the program do the check (or let makedist fail for invalid values). A typo in the name is probably more likely than the wrong order. If you check you'll have to update the script when the values change.
```
arg_ending=${1:(-7)}
# echo $arg_ending
if [[ $arg_ending == "-x86_64" || $arg_ending == "ppc64le" ]]; then
	echo "ERROR: you added a platform (...$arg_ending) where a source is required"
    usage
    exit 1
fi
```
After this the arguments should be OK

mambelli (changes requested):

creation/create_cvmfsexec_distros.sh
```diff
@@ -3,123 +3,185 @@
...
 
+####################### MAIN SCRIPT STARTS FROM HERE #######################
 
-# TODO: store/update version information in the $cvmfsexec_tarballs location for future reconfig/upgrade
-echo "$latest_ver" > "$cvmfsexec_tarballs"/.cvmfsexec_version
+# parsing the command-line arguments
+# check whether the first argument passed is the option '--work-dir'
+if [[ $1 == "--work-dir" ]]; then
+	# if value after this option is not a path (it is empty/unspecified), show usage documentation and exit
+	[[ ! -d "$2" ]] && usage || true
```
mambelli, line 150: Add an error message
```
if	[[ ! -d "$2" ]]; then
  echo "ERROR: the argument of --work-dir must be an existing directory
  usage
  exit 1
fi
```

mambelli (changes requested):

creation/create_cvmfsexec_distros.sh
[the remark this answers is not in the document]
mambelli: Yes, it is possible. If you use getopts it will do some checks for you, otherwise, you have to do them as you did checking that is a directory.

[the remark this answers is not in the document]
namrathaurs: Thanks for the clarification. Given the possibility that could happen, I've retained the logic that handles such a scenario.

creation/create_cvmfsexec_distros.sh
namrathaurs: updated as per this and the clarification discussion we had this week.

creation/create_cvmfsexec_distros.sh
namrathaurs: updated to reflect this

creation/create_cvmfsexec_distros.sh
```diff
@@ -3,123 +3,196 @@
...
+	echo ""
+	echo "DIR: full, absolute path to the factory work directory"
+	echo ""
+	echo "SOURCES_LIST (required): specifies the source(s) to download the latest"
+	echo "cvmfs configuration and repositories from. Must be at least one value "
+	echo "or a comma-separated list of values from the options {osg|egi|default}."
+	echo ""
+	echo "PLATFORMS_LIST (optional): indicates machine types (platform- and architecture-based)"
+	echo "for which distributions is to be built. Can be empty, a single value or a "
+	echo "comma-separated list of values from the options {rhel7-x86_64|rhel8-x86_64|suse15-x86_64}."
```
mambelli, line 26: You may consider heredocs when printing many lines. 
```
cat << EOF
Usage: $0 [--work-dir DIR] SOURCES_LIST [PLATFORMS_LIST]
...
EOF
```

creation/create_cvmfsexec_distros.sh
```diff
@@ -3,123 +3,196 @@
...
-    rm -rf "$cvmfsexec_latest"
+# parsing the command-line arguments
+# check whether the first argument passed is the option '--work-dir'
+if [[ $1 == "--work-dir" && ! -d "$2" ]]; then
+	# if value after this option is not a path (it is empty/unspecified), invoke the error handler
+	error_handler "The value of --work-dir option must be an existing directory."
+elif [[ -d "$1" ]]; then
+	# check if there was a path specified without the --work-dir option, invoke the error handler
+	error_handler "--work-dir option should precede directory path."
+fi
```
mambelli, line 161: This may cause some difficult error if there are directories named like the sources (osg, default, egi). I would skip this check and trust the following checks to catch a problem

creation/create_cvmfsexec_distros.sh
```diff
@@ -3,123 +3,196 @@
...
+	work_dir="$2"
+	shift 2
+else
+	# if --work-dir is not passed, assume default work-dir (RPM install)
+	work_dir="$DEFAULT_WORK_DIR"
 fi
 
+# check whether there are any remaining arguments passed to the script after the option
+if [[ $# -eq 0 ]]; then
+	echo -e "\nNo sources specified. Building/Rebuilding of cvmfsexec distributions disabled!"
```
mambelli, line 173: I see many `echo -e "\n..."`. It is OK not to add empty lines ahead of messages and previous messages should end w/ newline so this will start on its own line.

mambelli (changes requested):
Please add a line also to CHANGELOG.md

creation/create_cvmfsexec_distros.sh
namrathaurs: updated, thanks

creation/create_cvmfsexec_distros.sh
namrathaurs: This condition is purely checking if a directory path was entered without the `--work-dir` option preceding the path value and not the name of the directory. So, its unclear to me as to how this would cause the kinds of error that you described. Maybe I am not understanding your comment correctly?

creation/create_cvmfsexec_distros.sh
namrathaurs: This has been updated. The reason newline character was added to some of the `echo` statements was to be able to print the message on a newline, which was not happening without the newline character at the beginning of the message. To change this, the `factory_initd_startup_template` and `factory_initd_startup_template_sl7` scripts in `creation/template` were also modified.

creation/create_cvmfsexec_distros.sh
mambelli: Let's say I don't specify --work-dir (I want to use the default), so I call  `create_cvmfsexec_distros.sh osg` to build all platforms for osg, but I have locally a subdirectory called "osg". `[[ -d "$1" ]]` is true and I may not understand well what's happening because I forgot about that directory and I thought I was specifying a source.

mambelli:
```
$0 [--work-dir DIR] SOURCES_LIST [PLATFORMS_LIST]
```
If there is no --work-dir, $1 is a SOURCES_LIST (first argument) that is verified right below. The error "invalid source $1" and the usage string printed will be clear enough to understand that --work-dir needs to be added

namrathaurs: Discussed on Slack and agreed to incorporate this suggestion. The check has been removed since the check for sources will cover the scenario that I described in my response to the original comment in this thread.

doc/factory/configuration.html
```diff
@@ -964,6 +967,55 @@ <h3 class="western"><a name="args_global"></a>Global arguments</h3>
...
+              platforms=&quot;<i>platforms_list</i>&quot; /&gt;
+            </div>
+
+            <p>
+              This knob provides control over the building and/or rebuilding of
+              cvmfsexec distributions when CVMFS is to be provisioned on demand.
+              Distributions built via this knob will be located in factory work
+              directory location (by default, for an RPM-based install of
+              GlideinWMS, this would typically be
+              <i>/var/lib/gwms-factory/work-dir</i>). Two attributes are
```
mambelli, line 984: Somewhere say the name of the subdirectory with the distributions

doc/factory/configuration.html
```diff
@@ -964,6 +967,55 @@ <h3 class="western"><a name="args_global"></a>Global arguments</h3>
...
+              <i>rhel8-x86_64</i> | <i>suse15-x86_64</i> }.
+            </p>
+            <p>
+              Note that, <b><i>sources</i></b> is a <b>required</b> attribute
+              and <b><i>platforms</i></b> is an <b>optional</b> attribute.
+              <li>
+                if <b><i>platforms</i></b> is empty and <b><i>sources</i></b> is
+                non-empty), the attribute value for <b><i>platforms</i></b> will
+                default to the string
+                <i>rhel7-x86_64,rhel8-x86_64,suse15-x86_64</i>.
```
mambelli, line 1009: Tell that the default is all available platforms:
```
... is non-empty), the distribution will be built for all available platforms (<i>rhel7-x86_64,rhel8-x86_64,suse15-x86_64</i>).
```

mambelli (changes requested):

doc/factory/configuration.html
namrathaurs: updated to include this detail

doc/factory/configuration.html
namrathaurs: rephrased as suggested.

mambelli (approved):

PR #278, opened 2023-03-20
Add PyYAML to the factory dependencies

mmascher:
In order to get `OSG_autoconf` to work this has been installed in the factory:

```
PyYAML.x86_64                        3.10-11.el7                    @base 
```

I think it should be added as a dependency to the factory RPM.

PS: I have not tested this PR, hopefully I got it right. Please double check.

### DIFF
--- a/build/packaging/rpm/glideinwms.spec
+++ b/build/packaging/rpm/glideinwms.spec
@@ -261,6 +261,7 @@ Requires: python36-requests
 Requires: python36-jwt
 %endif
 Requires: python3-rrdtool
+Requires: PyYAML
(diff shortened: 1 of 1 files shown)


mambelli (changes requested):
This should change to include the different names for SL7/EL8 as discussed in the meeting.
If the dependency is not needed because of the lib dependency, please close the PR w/o merging

mambelli (changes requested):
The packages are different for EL7/8+9

mambelli:
Marco Mascheroni verified and the dependency is OK. Can be closed

PR #279, opened 2023-03-24
Fix nullglob problems

mambelli:
Fixing possible nullglob problems:
- when using wildcards in a for loop, e.g. `for i in *.txt`, if there is no match the shell will return the string with wildcards as value in the loop
- to solve you have to check for file existence (`[ -e "$i" ] || continue`) or use `shopt -s nullglob` to change the behavior in bash

This PR upgrades also the isort version in pre-commit to 5.11.5 to fix an isort bug w/ poetry and maintain python 3.7 compatibility


namrathaurs (approved):
Changes look good, ready for merging!

PR #301, opened 2023-05-29
Removing wrong classad Python module from requirements.txt

mambelli:
Removing classad from requirements.txt as suggested by the HTCondor team (J.Patton, C.Bollig)

classad was pulling a pure Python implementation from  https://pypi.org/project/classad

Regular installations use the `python3-condor` RPM instead of PyPI.


namrathaurs (approved):
Ready to be merged!

PR #302, opened 2023-05-30
Deprecation of getchildren() in ElementTree

### LINKED ISSUE #299: Frontend config error on python 3.9. with alma 9
Following are the details as reported by Steven Timm:

```
Info: Uploading facts for hepcintfe01.fnal.gov to ssi-puppet.fnal.gov
[root@hepcintfe01 ~]# gwms-frontend upgrade
Using default frontend config file: /etc/gwms-frontend/frontend.xml
All connected Factories are at least v3.4.1
NOTE: Python 3 has stricter type restrictions which may cause match expressions to fail in execution time. Please, make sure match_attr types are appropriately defined.
  File "/sbin/reconfig_frontend", line 196, in <module>
    p3_compatible, report = check_python3_expr.main(xml, False, True)
  File "/usr/lib/python3.9/site-packages/glideinwms/creation/lib/check_python3_expr.py", line 212, in main
    for element in findall_path(tree.getroot(), "match"):
  File "/usr/lib/python3.9/site-packages/glideinwms/creation/lib/check_python3_expr.py", line 126, in findall_path
    for child in element.data.getchildren():
AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'getchildren'
Upgrading the frontend                                     [FAILED]
```

The suspicion was presumably that the frontend does not like one of the MatchExprs in the frontend.xml, but the error message does not clearly state which MatchExpr might be troublesome. If it is that, how can the troublesome MatchExpr be identified? (The frontend.xml file was attached via email to glideinwms-support mailing list).

namrathaurs:
Fixes #299.

A [quick search](https://github.com/ocrmypdf/OCRmyPDF/issues/583#issuecomment-647787355) of the `AttributeError` revealed that `getchildren()` method has been deprecated since Python 3.2 but has not been warned about until v3.8. Further, the deprecation has been removed in Python 3.9. [This was also suspected by Bruno Coimbra during a discussion]

Ref: [Python 3.8 documentation](https://docs.python.org/3.8/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.getchildren)

As part of the investigation, also checked to compare the deprecated behavior of `getchildren()` [from Python 3.6] with `list()` from v3.8+ to make sure the outputs were identical (i.e. the result set and the type of the result). The outputs, along with their type, from the deprecated and the current approaches seems to be consistent with one another.
(Thanks to @BrunoCoimbra for recommending to do this exercise as part of the investigation!)

Following is the code snippet from the testing: ()
```
[root@fermicloud512 gwms-frontend]# python3
Python 3.6.8 (default, Nov 10 2020, 07:30:01) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import xml.etree.ElementTree as ET
>>> a = ET.Element('a')
>>> b = ET.SubElement(a, 'b')
>>> c = ET.SubElement(a, 'c')
>>> d = ET.SubElement(a, 'd')
>>> e = ET.SubElement(b, 'e')
>>> f = ET.SubElement(d, 'f')
>>> g = ET.SubElement(d, 'g')
>>> for child in a.getchildren():
...     print(child)
... 
<Element 'b' at 0x7f29cf306f98>
<Element 'c' at 0x7f29cf28b958>
<Element 'd' at 0x7f29cf28b2c8>
>>> for child in a.getchildren():
...     print(type(child))
... 
<class 'xml.etree.ElementTree.Element'>
<class 'xml.etree.ElementTree.Element'>
<class 'xml.etree.ElementTree.Element'>
>>> for child in list(a):
...     print(child)
... 
<Element 'b' at 0x7f29cf306f98>
<Element 'c' at 0x7f29cf28b958>
<Element 'd' at 0x7f29cf28b2c8>
>>> for child in list(a):
...     print(type(child))
... 
<class 'xml.etree.ElementTree.Element'>
<class 'xml.etree.ElementTree.Element'>
<class 'xml.etree.ElementTree.Element'>
>>> 
```

### DIFF
--- a/creation/lib/check_python3_expr.py
+++ b/creation/lib/check_python3_expr.py
@@ -123,7 +123,7 @@ def findall_path(root, tag, elements=[]):
     else:
         element = root
 
-    for child in element.data.getchildren():
+    for child in list(element.data):
         newElement = SimpleNamespace()
         newElement.data = child
         newElement.parent = element
@@ -165,7 +165,7 @@ def match_attrs_to_tuples(match_attrs):
     """
 
     tuples = []
-    for attr in match_attrs.getchildren():
+    for attr in list(match_attrs):
         tuples.append((attr.attrib["name"], attr.attrib["type"]))
     return tuples
 


namrathaurs:
To fix the `AttributeError`, that is caused due to the deprecation of `getchildren()` method in the _xml.etree.ElementTree_ module, on a system with Python 3.9, you need to patch the Frontend and Factory running GlideinWMS v3.10.2 with the file from this PR.

Frontend: replace `/usr/lib/python3.6/site-packages/glideinwms/creation/lib/check_python3_expr.py` with the file [here](https://raw.githubusercontent.com/namrathaurs/glideinwms/i299/getchildren_deprecation/creation/lib/check_python3_expr.py).
Factory: replace `/usr/lib/python3.6/site-packages/glideinwms/creation/lib/check_python3_expr.py` with the file [here](https://raw.githubusercontent.com/namrathaurs/glideinwms/i299/getchildren_deprecation/creation/lib/check_python3_expr.py).

After replacing the files restart Factory and Frontend.

mambelli (approved):

PR #309, opened 2023-06-22
Avoid invocation of cvmfsexec distros building if not necessary

### LINKED ISSUE #308: Do not for call cvmfsexec script if not necessary
**Describe the bug**
Not really a bug, but when reconfiguring the ITB dev factory I got an out of memory. Turned out that since `subprocess` uses `fork` then the whole process memory is duplicated. And since the test machine is really small the memory is not enough.

I'd suggest to avoid doing the subprocess call if not necessary (cvmfs has been rutned off).

-   Stakeholders: CMS, FactoryOps
-   Components: factory

mmascher:
As identified in #308, this PR brings in a change to the existing use of `subprocess` module (via the GWMS `subprocessSupport` module) for efficiency purposes.

Currently, the `subprocessSupport` module is used to execute `/usr/bin/create_cvmfsexec_distros.sh` always, regardless of whether cvmfsexec distributions are required to be built or rebuilt. If CVMFS is not needed and the `cvmfsexec_distro` tag is not defined in the factory configuration (disabled mode), then calling `subprocessSupport` to execute the script is basically unnecessary. Additionally, since `subprocess` spawns a new process to execute the script, this might result in a memory-intensive operation in a production setting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Changes since the last release
 -   Removed `classad` from requirements.txt. The HTCSS team distributes only the `htcondor` library in PyPI which includes a different version of classad (PR #301)
 -   Fixing Python 3.9 deprecations (`imp`, `getchildren()` in `xml.etree.ElementTree`) (PR #302, PR #303)
 -   Populate missing Entry parameters for ARC CEs submissions (PR #304)
+-   Do not for call cvmfsexec script if not necessary (PR #309)
 
 ### Testing / Development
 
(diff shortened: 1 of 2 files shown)


namrathaurs:
@mmascher Upon reviewing this PR, I realized that the suggestion you made to `cgWParamDict.py` can be modified such that the invocation of `/usr/bin/create_cvmfsexec_distros.sh` can be added to the existing `if cfgs:` condition, which is right after the lines corresponding to the invocation of `subprocessSupport.iexe_cmd`, instead of adding an extra if condition. I modified accordingly and added a message to inform the user about the building/rebuilding being disabled since that was previously being handled inside the `create_cvmfsexec_distros.sh` script. I also modified the description you had originally added to the list of Bug Fixes in the CHANGELOG file. 

Thanks for catching this as well as sharing your suggestions for improvement!

namrathaurs (approved):
Added a few modifications to the suggested changes and the PR is ready for merging!

PR #312, opened 2023-07-05
Minor updates for on-demand cvmfsexec

namrathaurs:
This PR brings in a couple of updates to the existing on-demand cvmfs provisioning support in GlideinWMS. 

1. While `rhel9-x86_64` is supported as a machine type by _cvmfsexec_, the `makedist` script fails when trying to generate a cvmfsexec distribution for this machine type on an el7 platform which causes the factory reconfig/upgrade to fail. Since a majority of the worker nodes are on el7, the suggestion was to exclude `rhel9-x86_64` from our list of possible machine types until the _cvmfsexec_ package is able to handle this. CHANGELOG has been updated about generating an el9 distro on an el7 machine for cvmfsexec (see Known Issues section). 

2. The `cvmfsexec_distro` tag is now included in the factory configuration to be available out-of-the-box with a fresh install/upgrade of GWMS. Although previously, building of cvmfsexec distributions was disabled by default (even without this tag in the xml), the tag was added to enforce consistency.

### DIFF
--- a/creation/create_cvmfsexec_distros.sh
+++ b/creation/create_cvmfsexec_distros.sh
@@ -9,8 +9,9 @@
 # Hardcoded variables
 CVMFSEXEC_REPO="https://www.github.com/cvmfs/cvmfsexec.git"
 DEFAULT_WORK_DIR="/var/lib/gwms-factory/work-dir"
-# TODO: add rhel9, other rhel derivatives and opensuse derivatives to the default list below eventually
-DEFAULT_MACHINE_TYPES="rhel7-x86_64,rhel8-x86_64,suse15-x86_64"
+# TODO: periodically add rhel, suse and other derivatives as supported by cvmfsexec
+# NOTE: ignoring rhel9-x86_64 (although supported) since el7 tools cannot work with el9 files at the moment (as suggested by Dave Dykstra)
+DEFAULT_MACHINE_TYPES="rhel7-x86_64,rhel8-x86_64,suse15-x86_64,rhel8-aarch64,rhel8-ppc64le"
 
 usage() {
 cat << EOF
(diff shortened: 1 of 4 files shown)


creation/create_cvmfsexec_distros.sh
```diff
@@ -24,7 +25,7 @@ or a comma-separated list of values from the options {osg|egi|default}.
 
 PLATFORMS_LIST (optional): indicates machine types (platform- and architecture-based)
 for which distributions is to be built. Can be empty, a single value or a
-comma-separated list of values from the options {rhel7-x86_64|rhel8-x86_64|suse15-x86_64}.
+comma-separated list of values from the options {rhel7-x86_64|rhel8-x86_64|suse15-x86_64|rhel8-aarch64|rhel8-ppc64le}.
```
mambelli, line 28: In the instructions, there should be also rhel9

creation/create_cvmfsexec_distros.sh
```diff
@@ -177,8 +181,8 @@ fi
 
 # after confirming that the remaining number of arguments to be either 1 or 2
 re_sources="^((osg|egi|default),)*(osg|egi|default),?$"
-# TODO: update the regex for machine types upon checking with `makedist -h` periodically
-re_mtypes="^((rhel(7|8)|suse15)(-x86_64),?)+$"
+# TODO: update the regex for rhel9 and other machine types upon checking with `makedist -h` periodically
+re_mtypes="^((rhel(7|8)|suse15)(-(x86_64|aarch64|ppc64le),?))+$"
```
mambelli, line 185: Also here rhel9  should be in the allowed types. We are not blocking it, we are providing a warhing that it does not work in el7

mambelli (changes requested):

creation/create_cvmfsexec_distros.sh
namrathaurs: Updated as suggested.

**Note to self:** We still need to include el9 support for cases where cvmfsexec distributions is required to be generated on an el9 machine, even though doing so on an el7 machine fails (this detail has been added in the CHANGELOG suggesting to exercise caution when generating distributions for el9 on an el7 machine).

creation/create_cvmfsexec_distros.sh
namrathaurs: Updated as suggested.

mambelli (approved):

PR #317, opened 2023-07-06
Sort VOs to make sure generated xml is always the same in OSG_autoconf

mmascher:


mambelli (changes requested):
The unit test is failing (check gwms.LOCAL.pyunittest.d.test_OSG_autoconf.txt) I think because it has the old order. Please fix that.

mambelli (approved):

PR #320, opened 2023-07-11
Added debian11 as possible WN OS

mambelli:
Added also /etc/os-release as a possible source of information.
For now, it's secondary, it should be moved to primary.

### DIFF
--- a/creation/web_base/condor_platform_select.sh
+++ b/creation/web_base/condor_platform_select.sh
@@ -28,7 +28,7 @@ error_gen=$(gconfig_get ERROR_GEN_PATH "$glidein_config")
 condor_vars_file=$(gconfig_get CONDOR_VARS_FILE "$glidein_config")
 
 condor_os=$(gconfig_get CONDOR_OS "$glidein_config")
-if [ -z "$condor_os" ]; then
+if [[ -z "$condor_os" ]]; then
     condor_os="default"
 fi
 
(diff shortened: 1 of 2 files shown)


creation/web_base/condor_platform_select.sh
```diff
@@ -87,6 +88,41 @@ findversion_debian() {
   fi
 }
 
+findversion_os_release() {
+    local os_f=
+    local os_release dist_id dist_id_like ver major_ver
+    if [ -f "/etc/os-release" ]; then
```
namrathaurs, line 94: Would `[[` be better here? Or, any specific reason that single `[` was used instead?

creation/web_base/condor_platform_select.sh
```diff
@@ -87,6 +88,41 @@ findversion_debian() {
...
+        os_f="/etc/os-release"
+    elif [ -f "/usr/lib/os-release" ]; then
```
namrathaurs, line 96: Same comment as above

creation/web_base/condor_platform_select.sh
```diff
@@ -72,13 +72,14 @@ findversion_debian() {
   dist_id_line=$(grep "DISTRIB_ID" /etc/lsb-release)
   dist_rel_line=$(grep "DISTRIB_RELEASE" /etc/lsb-release)
   if [[ ${dist_id_line} == *"Debian"* ]]; then
-    [[ ${dist_rel_line:16:3} = "11." ]] && condor_os='linux-debian10' && return
+    [[ ${dist_rel_line:16:3} = "12." ]] && condor_os='linux-debian11' && return
```
namrathaurs, line 75: Should this be: `condor_os='linux-debian12'` (instead of `linux-debian11`) since `${dist_rel_line:16:3} = "12."` ?

namrathaurs (commented):
some general comments added

creation/web_base/condor_platform_select.sh
mambelli: this is on purpose because there is no condor tarball for 12

creation/web_base/condor_platform_select.sh
mambelli: done

PR #322, opened 2023-07-21
Fixed idtoken error

mambelli:
Fixed writing of cached token as str 
and removed in token_util.py extra decoding that was corrupting idtokens and causing the Glidein schedd authentication to fail.

jwt.encode can accept a string as a key, but will encode it in utf-8, not latin-1, causing a different value (corrupted token)

### DIFF
--- a/lib/token_util.py
+++ b/lib/token_util.py
@@ -8,6 +8,7 @@
 #   This is a collection of utility functions for HTCondor IDTOKEN generation
 
 
+import codecs
 import os
 import re
 import socket
@@ -77,12 +78,11 @@ def token_str_expired(token_str):
     return expired
 
 
-def simple_scramble(data):
+def simple_scramble(in_buf):
     """Undo the simple scramble of HTCondor
-    simply XOR with 0xdeadbeef
-    Using defaults.BINARY_ENCODING (latin-1) to have a 1-to-1 characted-byte correspondence
 
-    Source: https://github.com/CoffeaTeam/jhub/blob/master/charts/coffea-casa-jhub/files/hub/auth.py#L196-L235
+    simply XOR with 0xdeadbeef
+    Source: https://github.com/CoffeaTeam/coffea-casa/blob/master/charts/coffea-casa/files/hub-extra/auth.py
 
     Args:
         data(bytearray): binary string to be unscrambled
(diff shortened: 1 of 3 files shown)


lib/token_util.py
```diff
@@ -26,6 +27,11 @@
 from glideinwms.lib.subprocessSupport import iexe_cmd
 
 
+def byt(x):
+    # return x.encode(defaults.BINARY_ENCODING)
+    return codecs.latin_1_encode(x)[0]
```
namrathaurs, line 32: I see that this method has been defined newly, but is this to be used/invoked somewhere in this script instead of what was existing? Or, is this just a helper method for some other use case because I do not see this method used at all.

lib/token_util.py
```diff
@@ -77,56 +83,54 @@ def token_str_expired(token_str):
...
-            datum = struct.unpack("B", data[i])[0]
-        else:
-            datum = data[i]
-        rslt = datum ^ deadbeef[i % lbeef]
-        b1 = struct.pack("H", rslt)[0]
-        outb += ("%c" % b1).encode(defaults.BINARY_ENCODING)
-    return outb
+    out_buf = b""
+    for idx in range(len(in_buf)):
+        scramble = in_buf[idx] ^ deadbeef[idx % 4]  # 4 = len(deadbeef)
```
namrathaurs, line 101: Would it be better to write `len(deadbeef)` instead of a static value `4`?

lib/token_util.py
```diff
@@ -158,6 +161,22 @@ def sign_token(identity, issuer, kid, master_key, duration=None, scope=None):
...
+    #         payload: bytes,
+    #         key: AllowedPrivateKeys | str | bytes,
+    #         algorithm: str | None = "HS256",
+    #         headers: dict[str, Any] | None = None,
+    #         json_encoder: type[json.JSONEncoder] | None = None,
+    #         is_payload_detached: bool = False,
+    #         sort_headers: bool = True,
+    #     ) -> str:
+    # From source code: https://github.com/jpadilla/pyjwt/blob/72ad55f6d7041ae698dc0790a690804118be50fc/jwt/utils.py
+    # If master_key is str, then it is encoded w/ utf-8:  value.encode("utf-8")
```
namrathaurs, line 179: I see that these lines are commented. Is there any specific reason why these lines are retained instead of completely removing them?

namrathaurs (changes requested):
Added few comments which could be considered as suggestions if applicable/relevant.

lib/token_util.py
mambelli: To explain why bytes should be used or to use utf-8 if using str. Is it loo much, too long and confusing?

namrathaurs: > To explain why bytes should be used or to use utf-8 if using str. Is it loo much, too long and confusing?

Since I do not know the full technical context, it does seem a bit confusing because there are three URLs being referenced in the details regarding the `encode` method. Is it possible to describe the details concisely, given that the details seems rather important? Suggesting this because just looking at it the way it is does not make the detail quite obvious/evident.

lib/token_util.py
mambelli: removed

lib/token_util.py
mambelli: since DEADBEEF will never change I did it for performance reasons, adding the comment for clarity.
Should I still use len()? In the end si just a few usec.

lib/token_util.py
mambelli: I changed it to a shorter explanation which can be important for future changes of the code
```
    # master_key should be `bytes`. `str` could cause value changes if was decoded not using utf-8.
    # The manual (https://pyjwt.readthedocs.io/en/stable/api.html) is incorrect to list `str` only.
    # The source code (https://github.com/jpadilla/pyjwt/blob/72ad55f6d7041ae698dc0790a690804118be50fc/jwt/api_jws.py)
    # shows `AllowedPrivateKeys | str | bytes` and if it is str, then it is encoded w/ utf-8:  value.encode("utf-8")
```

namrathaurs (approved):
Changes look good, ready for merging!

PR #329, opened 2023-08-13
Updated github actions versions and fixed some references

mambelli:
Updated GitHub actions to use the latest version


namrathaurs (approved):
Changes can be merged with the main!

PR #334, opened 2023-08-16
Added support for timeout, encoding and logging to iexe_cmd()

mambelli:
Clarified also the documentation, e.g. adding the default encoding (utf-8) and exceptions raised

### DIFF
--- a/lib/subprocessSupport.py
+++ b/lib/subprocessSupport.py
@@ -7,15 +7,18 @@ import subprocess
 
 from subprocess import CalledProcessError
 
+from . import defaults
+
 # CalledProcessError(self, returncode, cmd, output=None, stderr=None)
 # Provides: cmd, returncode, stdout, stderr, output (same as stdout)
 # __str__ of this class is not printing the stdout in the error message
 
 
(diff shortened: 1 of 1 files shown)


lib/subprocessSupport.py
```diff
@@ -83,12 +101,29 @@ def iexe_cmd(cmd, useShell=False, stdin_data=None, child_env=None, text=True):
...
+            stdoutdata, stderrdata = process.communicate(input=stdin_data, timeout=timeout)
+        except subprocess.TimeoutExpired as e:
+            process.kill()
+            stdoutdata, stderrdata = process.communicate()
+            err_str = "Timeout running '{}'\nStdout:{}\nStderr:{}\nException subprocess.TimeoutExpired:{}".format(
+                cmd,
+                stdoutdata,
+                stderrdata,
+                e,
+            )
```
BrunoCoimbra, line 114: Could we use an fString (`f""`) here instead?

BrunoCoimbra (approved):
I made a change suggestion but found no show-stoppers. Good to go!

lib/subprocessSupport.py
mambelli: This was done by the autoformatter to shorten the line (<120)

PR #335, opened 2023-08-17
Fixed node validation and added case insensitive gconfig_get

mambelli:
Fixed node validation, was not reading the glidein_config due to a wrong variable name
This was noted by @namrathaurs in PR #245. Moving it here because is a bug that needs to be fixed in the main branch

Added also a case-insensitive option for gconfig_get()
gconfig_get VAR FILE GREP_OPTIONS
e.g. `gconfig_get GLIDEIN_Report_Failed "${glidein_config}" "-i"`
And I fixed some files that were still using grep+cut because of case-insensitivity


namrathaurs (approved):
Changes look good and are ready for merging.

PR #336, opened 2023-08-18
Ensure that credentials have absfname or generator attributes

BrunoCoimbra:
This PR prevents the `reconfig` and `upgrade` operations from accepting credentials without `absfname` or `generator` attributes. This is a temporary solution that will be deprecated when we finish refactoring the credentials class.

**NOTE 1**
_The problem was caused by checking `if pel["absfname"] is None` instead of `if not pel["absfname"]`. There are other statements on the same function using the same syntax. We should consider updating them before merging this PR._

**NOTE 2**
_The changes on this PR access `if pel["absfname"].data` directly. I don't like the idea of accessing the internals of a `VOFrontendSubParams` object directly, but this might be an acceptable solution while we refactor the credentials class. Feel free to suggest a better way to update the object for now._

### DIFF
--- a/creation/lib/cvWParamDict.py
+++ b/creation/lib/cvWParamDict.py
@@ -1159,13 +1159,14 @@ def populate_common_descript(descript_dict, params):
         # print params.security.credentials
         for pel in params.security.credentials:
             validate_credential_type(pel["type"])
-            if not pel["absfname"]:
-                if not pel["generator"]:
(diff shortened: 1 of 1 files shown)


mambelli:
I added a local variable to avoid changing the value of the SubParam "absfname".

mambelli (approved):

Issue #338, opened 2023-08-23
Missing monitoring and glidefactoryclient ClassAd on Alma9


mambelli:
**Describe the bug**
When running on EL9 (Alma Linux 9) there is no monitoring information (RRD files) on the Factory and there are no glidefactoryclient ClassAds. The other effect is that the Factory continued to submit pilots since it is not aware of the submitted ones.

**To Reproduce**
Install a Factory on Alma9, run 3.10.3 RC1 (probably also earlier versions)

**Expected behavior**
RRD files on the Factory and glidefactoryclient ClassAds should be there.

**Screenshots**
NA

**Info (please complete the following information):**
Stakeholders and components can be a comma separated list or on multiple lines.
If you add a new stakeholder or component, not on the sample list, add it on a line by its own.

-   GlideinWMS version: 3.10.3 RC1
-   Python version: 3.9
-   OS version: EL9
-   HTCondor version: 9.0.17
-   Priority: high
-   Stakeholders: 
-   Components: factory, factory monitoring

**Additional context**
The Factory group log file contains information about a possible Python interpreter crash (memory "double free")  
Bruno and Namratha have been troubleshooting this and isolating to a problem in the RRD binding invocation.
`rrdSuppoty.py` can work both with binding or invoking the tools. Commenting the binding import fixes the issue but from previous experience, it is not scalable.

namrathaurs:
Adding detailed notes from the investigation:

As described in the issue description, along with the glidefactoryclient classad missing and no monitoring information available (rrd files) in `/var/lib/gwms-factory/work-dir/monitor/entry_ITB_CE_EL9_SciToken`, the following lines were found in `/var/log/gwms-factory/server/factory/group_0.err.log`:
```
[2023-09-12 16:44:32,714] DEBUG: glideFactoryEntryGroup:357: All children forked for glideFactoryEntry.check_and_perform_work terminated - took 0.07165884971618652 seconds. Loading post work state for the entry.
[2023-09-12 16:44:32,714] DEBUG: glideFactoryEntryGroup:368: No work found for entry ITB_CE_EL9_SciToken from any frontends
[2023-09-12 16:44:32,714] DEBUG: glideFactoryEntryGroup:378: Unable to process response from one or more children for check_and_perform_work. One or more forked processes may have failed and may not have client_stats updated
[...]

PR #339, opened 2023-08-31
Fix for rrdtool bindings crash when file is missing in EL9

namrathaurs:
**Partially** fixes issue #338.

Title was: "Workaround for the existing handling of rrdtool bindings", but this code fixes the problem.
A workaround would have been to disable the bindings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ Changes since the last release
 -   Populate missing Entry parameters for ARC CEs submissions (PR #304)
 -   Modified the usage of subprocess module, for building/rebuilding cvmfsexec distributions, only when necessary (PR #309)
 -   manual_glidein_submit now correctly sets idtokens in the EncryptedInputFiles (issue #283, PR #284)
+-   Fixed fetch_rrd crash in EL9 causing missing monitoring and glidefactoryclient classad information (Issue #338, PR #339)
 
 ### Testing / Development
 
@@ -38,8 +39,6 @@ Changes since the last release
 
 -   When generating cvmfsexec distribution for el9 machine type on an el7 machine, the factory reconfig and/or upgrade fails as a result of an error in `create_cvmfsexec_distros.sh`. This is possibly due to the tools for el7 being unable to handle el9 files (as per Dave Dykstra). Please exercise caution if using `rhel9-x86_64` in the `mtypes` list for the `cvmfsexec_distro` tag in factory configuration.
     -   Our workaround: the el9 machine type has been removed from the default list of machine types supported by the custom distros creation script (PR #312)
--   Installing the factory on Alma9 results in missing monitoring and glidefactoryclient classad. As a consequence of this, the Factory continued to submit pilots since it was not aware of the submitted pilots. See Issue #338 for the technical details.
-    -   Our workaround: added a check to test if the rrd file under question exists. If the file is non-existent, fetching the rrd file will be skipped.
 
 ## v3.10.2 \[2023-5-10\]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ Changes since the last release
 
 -   When generating cvmfsexec distribution for el9 machine type on an el7 machine, the factory reconfig and/or upgrade fails as a result of an error in `create_cvmfsexec_distros.sh`. This is possibly due to the tools for el7 being unable to handle el9 files (as per Dave Dykstra). Please exercise caution if using `rhel9-x86_64` in the `mtypes` list for the `cvmfsexec_distro` tag in factory configuration.
     -   Our workaround: the el9 machine type has been removed from the default list of machine types supported by the custom distros creation script (PR #312)
+-   Installing the factory on Alma9 results in missing monitoring and glidefactoryclient classad. As a consequence of this, the Factory continued to submit pilots since it was not aware of the submitted pilots. See Issue #338 for the technical details.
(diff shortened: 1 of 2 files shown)


lib/rrdSupport.py
```diff
@@ -464,7 +464,10 @@ def fetch_rrd(self, filename, CF, resolution=None, start=None, end=None, daemon=
             args.append("--daemon")
             args.append(str(daemon))
 
-        return self.rrd_obj.fetch(*args)
+        if os.path.exists(filename):
+            return self.rrd_obj.fetch(*args)
+        else:
+            raise FileNotFoundError(f"File {filename} does not exist. Skipping...")
```
mambelli, line 470: Even if FileNotFoundError is more correct, I'd raise  RuntimeError which has been normally used to indicate problems w/ something in the library. Or check what is caught upstream.
```
            raise RuntimeError(f"RRD File '{filename}' does not exist. Failing fetch_rrd.")
```

CHANGELOG.md
```diff
@@ -37,6 +37,8 @@ Changes since the last release
 
 -   When generating cvmfsexec distribution for el9 machine type on an el7 machine, the factory reconfig and/or upgrade fails as a result of an error in `create_cvmfsexec_distros.sh`. This is possibly due to the tools for el7 being unable to handle el9 files (as per Dave Dykstra). Please exercise caution if using `rhel9-x86_64` in the `mtypes` list for the `cvmfsexec_distro` tag in factory configuration.
     -   Our workaround: the el9 machine type has been removed from the default list of machine types supported by the custom distros creation script (PR #312)
+-   Installing the factory on Alma9 results in missing monitoring and glidefactoryclient classad. As a consequence of this, the Factory continued to submit pilots since it was not aware of the submitted pilots. See Issue #338 for the technical details.
+    -   Our workaround: added a check to test if the rrd file under question exists. If the file is non-existent, fetching the rrd file will be skipped.
```
mambelli, line 41: Instead of these known issues add only a Bug Fix
```
-  Fixed fetch_rrd crash in EL9 causing  missing monitoring and glidefactoryclient classad (Issue #339, PR #339)
```
You can add the details here in the PR or Issue

mambelli (changes requested):

lib/rrdSupport.py
namrathaurs: Checking to compare against the behavior in EL7 was a great suggestion and an eye-opening one! 

As per your suggestion, I used the code from the `rrdtool_crash.py` test script that was written during the troublehsooting of the issue #338 and executed the Python code in interpreter mode. Following is the output on EL7 and Alma9 platforms, respectively:
```
# EL7 - running rrdtool.x86_64 1.8.0-1.2.osg36.el7
>>> import rrdtool
>>> rrdo = rrdtool
# intentionally used the wrong file name to mimic the behavior of a non-existent file
>>> static_args = ['/var/lib/gwms-factory/work-dir/monitor/entry_fc445_ce-new/frontend_vofrontend_service_frontend/Stauts_Attributes.rrd', 'AVERAGE', '-r', '300', '-e', '1693520700', '-s', '1693513500']
>>> result = rrdo.fetch(*static_args)
rrdtool.OperationalError: opening '/var/lib/gwms-factory/work-dir/monitor/entry_ITB_CE_EL9_SciToken/frontend_vofrontend_service_frontend/Stauts_Attributes.rrd': No such file or directory
```

```
# EL9 - running rrdtool.x86_64 1.7.2-21.el9
>>> import rrdtool
>>> rrdo = rrdtool
>>> static_args = ['/var/lib/gwms-factory/work-dir/monitor/entry_fc445_ce-new/frontend_vofrontend_service_frontend/Status_Attributes.rrd', 'AVERAGE', '-r', '300', '-e', '1693520700', '-s', '1693513500']
>>> result = rrdo.fetch(*static_args)
double free or corruption (out)
Aborted (core dumped)
```

With v1.8 coming from `osg36` repo on EL7, rrdtool throws a completely specific exception `rrdtool.OperationalError` when there is a non-existent file, which does not seem to be crashing the Python interpreter. On the other hand, with v1.7 coming from `appstream` repo on Alma9, the exception thrown appears to be crashing the Python interpreter level with the message `double free or corruption (out)`. Additionally, when the `rrdtool_crash.py` was run at the command line as `python rrdtool_crash.py`, the message returned was `Segmentation fault (core dumped)` which basically implied that the Python interpreter crashed.

@mambelli, Do you think using `OperationalError` to raise the error instead of RuntimeError would be better here?

mambelli: I still prefer RuntimeError. Or if you really want I'm OK with FileNotFoundError (checking the callers), at least these are both standard python exceptions and a bit more limited than Exception.
Both `rrdtool.OperationalError` (errors generated by rrdtool itself) and `rrdtool.ProgrammingError` (programming errors, e.g. passing wrong datatypes) are derived directly from `Exception` and custom to rrdtool.
I'd be for leaving these inside the rrdSupport.py module so that rrdtool does not have to be imported in other places.
If desired the rrdtool exceptions could be catched and chained to another raised exception.

CHANGELOG.md
namrathaurs: Updated as a bug fix instead of being a known issue as suggested. Will add more details in the PR/Issue shortly, including the test script that was used as part of the troubleshooting work.

lib/rrdSupport.py
namrathaurs: This has been taken into consideration and the latest commit addresses this.

mambelli (approved):

PR #340, opened 2023-09-03
ready for release v3_10_3_rc2

mambelli:
Release branch
Tags edited.
Fixes spec file


namrathaurs (changes requested):
In `doc/history.html` file, there are two mentions of the bug fix concerning `manual_glidein_submit`, which to me, looks like a repetition. However, the PR number in the second occurrence is different from the first mention of `manual_glidein_submit`. Are these two separate parts of the same bug fix perhaps

(Github did not allow me to add a comment directly against the line in the file under question, so left a general comment here instead)

mambelli:
It was a duplicate. Fixed both CHANGELOG.md and various release notes.

mambelli:
Completed requested change. Merging

PR #350, opened 2023-09-13
gconfig.py should not use python3 but gwms-python. Fixes #349

### LINKED ISSUE #349: gconfig.py should not use python3 but gwms-python
**Describe the bug**
Some sites do not have python3 and the gconfig.poy fails when it is executed as a validation script since it has python3 in the shebang.

**Expected behavior**
We should use the gwms-python wrapper for the shebang so that the script works on all sites

**Screenshots**
```
Executing (flags:) /data/glide_1anBib/main/gconfig.py
/usr/bin/env: python3: No such file or directory
=== Validation error in /data/glide_1anBib/main/gconfig.py ===
Wed Sep 13 17:10:37 CEST 2023 Error running '/data/glide_1anBib/main/gconfig.py'
     The test script did not produce an XML file. No further information available.
Wed Sep 13 17:10:37 CEST 2023 Notifying VO of error
Removed condor variables
Advertising failure to the VO collector
```

-   GlideinWMS version: 3.10.3
-   Priority: high
-   Stakeholders: FactoryOps
-   Components: glidein

mmascher:
There may still be sites with only python/python2, so `env python3` may fail.
`gwms-python` should be used.
Fixes #349

### DIFF
--- a/creation/web_base/gconfig.py
+++ b/creation/web_base/gconfig.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!main/gwms-python
 
 # SPDX-FileCopyrightText: 2009 Fermi Research Alliance, LLC
 # SPDX-License-Identifier: Apache-2.0
--- a/creation/web_base/gconfig.py
+++ b/creation/web_base/gconfig.py
@@ -1,9 +1,24 @@
-#!main/gwms-python
+#!/bin/sh
 
 # SPDX-FileCopyrightText: 2009 Fermi Research Alliance, LLC
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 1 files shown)


mambelli:
I actually started a PR with a similar solution but then I thought that the script is launched during the setup (to test it) and later is designed to be mostly imported but could be also launched and needs to find the Python interpreter.
I found this solution on https://stackoverflow.com/a/60696126/3547046
There are also interesting discussions on using `type`, `command -v`, etc: 
- https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script
- (with nice history background) https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then

mambelli (approved):

PR #353, opened 2023-09-14
Fixed alt code still using python (ie python2) interpreter

mambelli:
Fixed alt code still using python (i.e. python2) interpreter
There were some alternative routines (for jq in logging) and for base64  in a couple of places, using python code when the shell utility was not available.
Now the code is compatible w/ both python2 and python3 and the invocation also looks first for python3 and then for python2.


namrathaurs (approved):
Changes are ready for merging!

PR #355, opened 2023-09-19
Workaround for EL7 PyJWT generating bytes instead of str

mambelli:
There is a bug in older PyJWT versions, like the one in EL7. `jwt.encode()` is returning bytes instead of str. 
Here I'm adding a workaround to protect against it and work with both new and older versions of PyJWT

Here is the PyJWT issue: https://github.com/jpadilla/pyjwt/issues/391
This was fixed in PyJWT v2.0.0, https://github.com/jpadilla/pyjwt/pull/513

### DIFF
--- a/factory/glideFactory.py
+++ b/factory/glideFactory.py
@@ -233,7 +233,7 @@ def generate_log_tokens(startup_dir, glideinDescript):
             # Obtain a legal filename from the url, escaping "/" and other tricky symbols
             recipient_safe_url = urllib.parse.quote(recipient_url, "")
 
-            # Generate the token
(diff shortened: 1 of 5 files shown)


mambelli:
To Patch an existing EL7 installation, replace `/usr/lib/python3.6/site-packages/glideinwms/lib/token_util.py`  with the one form this PR,  https://raw.githubusercontent.com/mambelli/glideinwms/v310/pyjwt_encode_bug_workaround/lib/token_util.py and restart the Frontend with `systemctl restart gwms-frontend`

BrunoCoimbra (approved):

PR #359, opened 2023-09-27
Fixed TRUST_DOMAIN in setup_x509 

mambelli:
Fixed setup_x509 to be successful w/ TRUST_DOMAIN set in the as Factory or Frontend parameter

### DIFF
--- a/creation/web_base/setup_x509.sh
+++ b/creation/web_base/setup_x509.sh
@@ -441,7 +441,9 @@ _main() {
         gconfig_add GLIDEIN_CONDOR_TOKEN "$GLIDEIN_CONDOR_TOKEN"
         gconfig_add GLIDEIN_CONDOR_TOKEN_ORIG "$GLIDEIN_CONDOR_TOKEN_ORIG"
         out=$(gconfig_get TRUST_DOMAIN "${glidein_config}")
-        if [[ -z "$out" ]]; then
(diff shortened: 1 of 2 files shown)


creation/web_base/setup_x509.sh
```diff
@@ -441,7 +441,9 @@ _main() {
         gconfig_add GLIDEIN_CONDOR_TOKEN "$GLIDEIN_CONDOR_TOKEN"
         gconfig_add GLIDEIN_CONDOR_TOKEN_ORIG "$GLIDEIN_CONDOR_TOKEN_ORIG"
         out=$(gconfig_get TRUST_DOMAIN "${glidein_config}")
-        if [[ -z "$out" ]]; then
+        if [[ -n "$out" ]]; then
+            cred_updated+=idtoken,
```
namrathaurs, line 445: Is it intended to have the comma at the end of the line `cred_updated+=idtoken,`  ?

namrathaurs (commented):
Just a comment as I was not sure of the exact behavior/reason behind adding the comma if that's essential.

creation/web_base/setup_x509.sh
mambelli: yes, it has to, it is a list of values

mambelli:
resolved comments, merging

PR #360, opened 2023-09-27
GLIDEIN_SINGULARITY_BINARY_OVERRIDE set as parameter

mambelli:
Make so GLIDEIN_SINGULARITY_BINARY_OVERRIDE can be set also in the Frontend and Factory params and not only in the workernode environment

### DIFF
--- a/creation/web_base/singularity_lib.sh
+++ b/creation/web_base/singularity_lib.sh
@@ -406,7 +406,7 @@ gwms_from_config() {
     # 3. - function to validate or process the parameter (get_prop_bool or same interface)
     #      The default, when used, is not processed by this function
     local ret=
-    if [[ -n "$glidein_config" ]]; then
+    if [[ -r "$glidein_config" ]]; then
         #ret=$(grep "^$1 " "$glidein_config" | cut -d ' ' -f 2-)
         #ret=$(tac "$glidein_config" | grep -m1 "^$1 " | cut -d ' ' -f 2-)
         if ! ret=$(gconfig_get "$1" "$glidein_config" 2>/dev/null); then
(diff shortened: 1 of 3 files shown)


creation/web_base/singularity_lib.sh
```diff
@@ -1498,7 +1498,8 @@ singularity_locate_bin() {
     local test_out
     HAS_SINGULARITY=False
     local osg_singularity_binary="${OSG_SINGULARITY_BINARY:-${OSG_SINGULARITY_BINARY_DEFAULT}}"
-    local singularity_binary_override="${GLIDEIN_SINGULARITY_BINARY_OVERRIDE}"
+    local singularity_binary_override
+    singularity_binary_override="${GLIDEIN_SINGULARITY_BINARY_OVERRIDE:-$(gwms_from_config GLIDEIN_SINGULARITY_BINARY_OVERRIDE)}"
```
namrathaurs, line 1502: I do not see any reference to `gwms_from_config`, so wanted to make sure if this is set/imported in a script that invokes the method that has this line?

namrathaurs (commented):
Comment added. Also, I see that the bats test step failed in the Github workflow. I'm guessing that needs to be fixed before this can be merged?

creation/web_base/singularity_lib.sh
mambelli: It's above, https://github.com/glideinWMS/glideinwms/blob/f0065cca512e12ef0a55ede42af390edf8b9cb3c/creation/web_base/singularity_lib.sh#L401

mambelli:
Fixed test and addressed comments, merging

PR #365, opened 2023-10-11
Improvements in glidein startup script

namrathaurs:
Added some indentation and structural improvements to the existing `glidein_startup.sh` script to ensure consistency  and promote better readability of code, respectively.
Note that these changes do not modify the existing logic/behavior, so the underlying logic of the glidein startup process is preserved from before.


mambelli:
This has been superseded by #373 including all its changes.

PR #379, opened 2023-12-28
Fixed add_to_path() to avoid partial path matching

mambelli:
Fixed add_to_path() to avoid partial path matching. 
The function prunes from the PATH/GWMS_PATH occurrences of the path being added to avoid indefinite growth of paths with multiple occurrences of the same path.
Before the fix the pruning could have been removing a partial matches, e.g. `/this/path_long` would have been removed when adding `/this/path`.
Added clarification comments


namrathaurs (approved):
Changes approved and are ready for merging!

Issue #362, opened 2023-09-28
CERN production factory crashes


mmascher:
**Describe the bug**
The CERN producction factory crashed two times in the past week. It seems it was doing the rotation of the entry logfiles. No alarm was actually fired since the python processes keeps running.

**To Reproduce**
Hard. It just happens from time to time. Maybe increase the rotation frequency of the logs and observe?

**Screenshots**
```
[2023-09-27 14:49:59,951] DEBUG: cleanupSupport:37: Forked cleanup PIDS [123125, 123126, 123127, 123128]
[2023-09-27 14:56:55,733] DEBUG: glideFactoryEntryGroup:308: Setting parallel_workers limit of 8
[2023-09-27 15:00:56,094] WARNING: glideFactoryEntryGroup:415: Error occurred while trying to find and do work.
[2023-09-27 15:00:56,095] ERROR: glideFactoryEntryGroup:416: Exception: 
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/glideinwms/factory/glideFactoryEntryGroup.py", line 412, in iterate_one
    do_advertize, factory_in_downtime, glideinDescript, frontendDescript, group_name, my_entries
  File "/usr/lib/python3.6/site-packages/glideinwms/factory/glideFactoryEntryGroup.py", line 344, in find_and_perform_work
    logSupport.roll_all_logs()
  File "/usr/lib/python3.6/site-packages/glideinwms/lib/logSupport.py", line 289, in roll_all_logs
    handler.check_and_perform_rollover()
  File "/usr/lib/python3.6/site-packages/glideinwms/lib/logSupport.py", line 283, in check_and_perform_rollover
    if self.shouldRollover(None, empty_record=True):
  File "/usr/lib/python3.6/site-packages/glideinwms/lib/logSupport.py", line 186, in shouldRollover
    self.stream.seek(0, 2)  # due to non-posix-compliant Windows feature
ValueError: I/O operation on closed file.
[2023-09-27 15:00:56,215] DEBUG: glideFactoryEntryGroup:418: Group Work done: {}
```

**Info (please complete the following information):**

-   GlideinWMS version: 3.10.4
-   Priority: high
-   Stakeholders: CMS, FactoryOps
-   Components: factory

mmascher:
This happened again on Tuesday Novermber 2nd. It was a while.

mambelli:
[...]

PR #389, opened 2024-01-17
Workaround and extra logging to investigate bug in log rotation

mambelli:
Added workaround and extra logging to investigate a bug in log rotation
Added also some compatibility adaptors to run unit tests on a Mac

This improves issue #362

### DIFF
--- a/factory/glideFactoryEntryGroup.py
+++ b/factory/glideFactoryEntryGroup.py
@@ -13,7 +13,7 @@
 #   $3 = advertize_rate (int): The rate at which advertising should occur (every $3 loops)
 #   $4 = startup_dir (str|Path): The "home" directory for the entry.
 #   $5 = entry_names (str): Colon separated list with the names of the entries this process should work on
-#   $6 = group_id (str): Group id, normally a number (with the "group_" prefix formes the group name),
+#   $6 = group_id (str): Group id, normally a number (with the "group_" prefix it forms the group name),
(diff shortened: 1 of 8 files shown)


factory/glideFactoryEntryGroup.py
```diff
@@ -29,7 +29,7 @@
 from glideinwms.factory import glideFactoryLib as gfl
 from glideinwms.factory import glideFactoryPidLib
 from glideinwms.lib import classadSupport, cleanupSupport, logSupport
-from glideinwms.lib.fork import fetch_fork_result_list, ForkManager
+from glideinwms.lib.fork import fetch_fork_result_list, ForkManager, print_child_processes
```
mmascher, line 32: Isn't this a syntax error? I mean the comma after ForkManager?

mmascher (commented):
Everything else is good, just check the syntax error of the earlier comment.

factory/glideFactoryEntryGroup.py
mambelli: I don't think. And pylint and the unit test are OK. The comma should be after all items after the first one: https://docs.python.org/3/reference/simple_stmts.html#the-import-statement

mmascher: For some reason I saw `ForkManager` written in red and my brain registered it as the`import keyword`. all good.

mmascher (approved):

Issue #388, opened 2024-01-16
Factory crashing on update to 3.10.6 because submit_attrs.cfg is not created when there are no submit attributes


mmascher:
The actual error is that `submit_attrs.cfg` files are not created in the work-dir and entries subdirectories when upgrading an existing Factory and in absence of content to put in the files (no submit_attr entries for that section -- global or entry config). The missing files cause the Factory to crash.
See the comment below for a test and a workaround.

ISSUE EDITED. THIS ISSUE WAS:  "Entries might be configured with no submit attributes"

Still related to this https://github.com/glideinWMS/glideinwms/pull/382

After applying the fix from #387 I hit another road blocker. It turns out that not all the entries have submit attributes, and when an entry does not have any submit attribute a submit_attrs.cfg file is not created. In order to conrtinue with the tests in ITB what I did was to run this command:

```
[mmascher@vocms0205 work-dir]$ pwd
/var/lib/gwms-factory/work-dir
[mmascher@vocms0205 work-dir]$ for dir in `find . -maxdepth 1 -type d`; do sudo -u gfactory touch $dir/submit_attrs.cfg; done
```

The exception I was hitting was:

```
[...]

PR #391, opened 2024-01-19
Fixed writing of missing dict file during upgrade

mambelli:
Fixed writing of missing dict file during upgrade

Added and improved many docstrings

This fixes the symptom by creating the file when missing (i.e. things will work) but there is some code setting incorrectly the `self.changed` flag that should be fixed in the future. I added a TODO in the code.


mmascher (approved):

mambelli:
This fixes #388

mmascher (commented):
LGTM

PR #396, opened 2024-01-30
Fixing Apptainer output validation in apptainer initial test

### LINKED ISSUE #395: Hosted CE module loading / singularity detection weirdness
**Describe the bug**
A hosted CE is failing to detect apptainer.
From the detailed logs
[job.3610066.0.err.txt](https://github.com/glideinWMS/glideinwms/files/14091133/job.3610066.0.err.txt)
[job.3610066.0.out.txt](https://github.com/glideinWMS/glideinwms/files/14091134/job.3610066.0.out.txt)
, apptainer runs correctly (EC 0), but fails the validation of the output.
Possible causes are:
- Apptainer/Singularity test fails when /proc file system is not available in the image
- The format of /proc is different in the image used for the test

There could also be a problem with the image or Apptainer execution
This behavior requires further investigation

**To Reproduce**
Apptainer is in tmp/glide_Pb9H5L/main/singularity_setup.sh invoked w/ the command:
```
"/cm/local/apps/apptainer/current/bin/singularity" -vvv -d  exec --home "/tmp/glide_Pb9H5L":/srv --pwd /srv  --ipc --contain  --pid --bind "/etc/hosts,/etc/localtime"  "/tmp/glide_Pb9H5L/images/htc__rocky__9.sif" cat /proc/self/uid_map
```
And the result is:
```
INFO  Singularity at '/cm/local/apps/apptainer/current/bin/apptainer' failed (ec:0) w/ unexpected output
```

**Expected behavior**
It is unexpected to have EC 0 and not the correct output

**Screenshots**

-   GlideinWMS version: 3.10.5
-   Python version:
-   OS version:
-   HTCondor version:
-   Priority:  high
-   Stakeholders: FactoryOps, OSG
-   Components: glidein

**Additional context**

mambelli:
The normal singularity/apptainer validation has 2 parts:
1. return exit code 0
2. return a valid output that allows us to determine the apptainer/singularity mode or at least verify the correct execution of the command in the provided image

The 1st commit was "Accepting apptainer test also when returning the wrong output", was disabling 2 and accepting the apptainer executable as long as the exit code is 0. It also provides further debug output to help with troubleshooting. It was as a temporary workaround. 

The second commit is the final solution and fixes #395 
It kept the additional debug output, it is back to failing if the test returns the wrong output, and it fixes the regular expression validating the output.
There may be no initial blank if the first number in /proc/self/uid_map  is big enough. This was tailing the previous expression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,29 @@ SPDX-FileCopyrightText: 2009 Fermi Research Alliance, LLC
 SPDX-License-Identifier: Apache-2.0
 -->
 
+## Changes Since Last Release OR vX.Y.Z \[yyyy-mm-dd\]
+
+Changes since the last release
+
+### New features / functionalities
+
+-   item one of the list
+-   item N
+
+### Changed defaults / behaviours
+
+### Deprecated / removed options and commands
+
+### Security Related Fixes
+
+### Bug Fixes
+
+-   Fixed Apptainer validation not considering uid_map w/o initial blank (Issue #395, PR #396)
+
+### Testing / Development
+
+### Known Issues
+
 ## v3.10.6 \[2024-01-25\]
 
 Minor new features, mostly a bug fix release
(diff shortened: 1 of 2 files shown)


mmascher (approved):
LGTM

PR #406, opened 2024-05-01
Added min_memory requirement for usable partitionable slot Glideins

### LINKED ISSUE #405: Improve glidein requesting in cases where there is a high-memory job mix
DUNE typically runs with a 5.5GB RequestMemory.  This often causes the frontend to believe that we have many usable cores when in fact we don't because there is not enough memory to match another job.  Thus not enough glideins are requested.. to the extreme that we can have 12K jobs in the queue but only 2400 jobs running.

We would like to make the minimum free memory per glidein configurable.
Currently it is hard-wired to 2500 at this line of code.
https://github.com/glideinWMS/glideinwms/blob/01d534e9467a5f4496ba2828b902490f6966be99/frontend/glideinFrontendLib.py#L811
If this is configurable then we could change the configuration to adjust for different mixes of jobs.  But in the near term we expect that all of our jobs will be high-memory through our current beam run.
In general it might actually be nice to supply a configurable condor_status query in which the VO can determine for itself what slots are available and what slots are not.  This could allow for factors other than memory--some remote sites are short on disk too and that can affect glidein occupancy as well.

Marco has also suggested increasing the idle_vms_per_entry and idle_vms_total settings in the
configuration and we are trying that first.  if that doesn't work then we will hot-patch the line of code above.

- Priority:  high
- Stakeholders: DUNE
- Components:  Frontend

**Additional context**
Add any other context or supporting files about the feature request here.

Eventually the same may need to be done for the Decision Engine which has similar but not identical logic.

mambelli:
Added min_memory requirement to consider usable partitionable slot Glideins. Partitionable Glideins with less than min_memory MB of RAM (and their cores) are discarded.

Changes the selection in glidienFrontendLib, the configuration, and the documentation.
PartGlideinMinMemory is the key in the Frontend Group description file
group.config.partitionable_glidein.min_memory is the entry in the XML config file

Updated the CHANGELOG also with lines about committed PRs

Added DUNE stakeholder in the new issue/feature auto-labeling

This PR fixes #405


StevenCTimm (commented):
you have a typo in glideinFrontendElement.py, glidein is spelled wrong (glidien)

[5:34](https://hepcloud.slack.com/archives/DM1BJJXFH/p1714602890990519)
status_dict_idle = glideinFrontendLib.getIdleCondorStatus(self.status_dict, self.p_glidien_min_memory)
[5:34](https://hepcloud.slack.com/archives/DM1BJJXFH/p1714602895114039)
glidein is spelled wrong
[5:36](https://hepcloud.slack.com/archives/DM1BJJXFH/p1714602990927879)
ditto line 167, 2169, 2339, 2142
[5:36](https://hepcloud.slack.com/archives/DM1BJJXFH/p1714602996708469)
Otherwise it looks fine

mmascher (approved):
Looks good to me

PR #414, opened 2024-06-09
Release v3 10 7 rc2

mambelli:
RC2 for 3.10.7
Compared to RC1, we reverted the condor configuration changes that were causing authentication errors and added the HTCondor download improvements.


namrathaurs (approved):

PR #415, opened 2024-06-13
Setting blackhole mechanism only if enabled in the configuration

mambelli:
Setting the blackhole mechanism only if enabled in the configuration

Before there was an error when the attributes were not set in the frontend/factory configuration.
The last error here was preventing the Startd from starting:
```
06/13/24 06:14:14 (pid:240950) eth0 enabled Wake-on: no (raw: 0x00)
06/13/24 06:14:14 (pid:240950) Using network interface eth0 for hibernation
06/13/24 06:14:14 (pid:240950) my_popenv: Failed to exec /var/lib/condor/execute/dir_226360/glide_gQGfYg/main/condor/libexec/condor_power_state, errno=2 (No such file or directory)
06/13/24 06:14:14 (pid:240950) Initially invoking hibernation plugin '/var/lib/condor/execute/dir_226360/glide_gQGfYg/main/condor/libexec/condor_power_state ad'
06/13/24 06:14:14 (pid:240950) Failed to run hibernation plugin '/var/lib/condor/execute/dir_226360/glide_gQGfYg/main/condor/libexec/condor_power_state ad'
06/13/24 06:14:14 (pid:240950) Detected hibernation states:
06/13/24 06:14:17 (pid:240950) VM universe will be tested to check if it is available
06/13/24 06:14:17 (pid:240950) ERROR "Syntax error in START expression: '((TRUE && (SiteWMS_WN_Draining =?= False)) && (ifThenElse(isUndefined((RecentJobBusyTimeAvg < (1.0/)) && (RecentJobBusyTimeCount >= )), False, (RecentJobBusyTimeAvg < (1.0/)) && (RecentJobBusyTimeCount >= )) =?= False)) && ((True) && (True) && (True) && (True)) && (((GLIDEIN_ToRetire =?= UNDEFINED) || (CurrentTime < GLIDEIN_ToRetire))) && (True)'" at line 587 in file /var/lib/condor/execute/slot1/dir_2508072/userdir/build-ba6TSX/BUILD/condor-10.6.0/src/condor_startd.V6/util.cpp
```
You can see how the missing attributes (shell variables in condor_startup.sh) are causing a malformed START expression.

### DIFF
--- a/creation/web_base/condor_startup.sh
+++ b/creation/web_base/condor_startup.sh
@@ -74,6 +74,11 @@ error_gen=$(gconfig_get ERROR_GEN_PATH "$config_file")
 # Read the knobs coming from the frontend configuration for blackhole detection (GLIDEIN_BLACKHOLE_NUMJOBS and GLIDEIN_BLACKHOLE_RATE)
 glidein_blackhole_numjobs=$(gconfig_get GLIDEIN_BLACKHOLE_NUMJOBS "$config_file")
 glidein_blackhole_rate=$(gconfig_get GLIDEIN_BLACKHOLE_RATE "$config_file")
+if [[ -z "$GLIDEIN_BLACKHOLE_RATE" || "$GLIDEIN_BLACKHOLE_RATE" =~ ^0(\.0*)?$ ]]; then
+    use_blackhole_prevention=false
+else
+    use_blackhole_prevention=true
+fi
 
 glidein_startup_pid=$(gconfig_get GLIDEIN_STARTUP_PID "$config_file")
 # DO NOT USE PID FOR DAEMON NAMES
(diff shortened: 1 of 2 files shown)


creation/web_base/condor_startup.sh
```diff
@@ -760,26 +765,31 @@ EOF
...
             num_slots_for_shutdown_expr=1
-            #Blackhole calculation based on the parent startd stats (child slots will have the same stats).
-            cat >> "$CONDOR_CONFIG" <<EOF
+            # Blackhole calculation based on the parent startd stats (child slots will have the same stats).
+	    if $use_blackhole_prevention; then
+                cat >> "$CONDOR_CONFIG" <<EOF
 BLACKHOLE_TRIGGERED_P = (RecentJobBusyTimeAvg < (1.0/$glidein_blackhole_rate)) && (RecentJobBusyTimeCount >= $glidein_blackhole_numjobs)
 GLIDEIN_BLACKHOLE = ifThenElse(isUndefined(\$(BLACKHOLE_TRIGGERED_P)), False, \$(BLACKHOLE_TRIGGERED_P))
 EOF
+            fi
```
namrathaurs, line 774: Is it okay if `fi` is indented more than it's opening `if`? I know this is in a shell script and indentation does not matter. Asking this for merely consistency purposes.

creation/web_base/condor_startup.sh
```diff
@@ -74,6 +74,11 @@ error_gen=$(gconfig_get ERROR_GEN_PATH "$config_file")
 # Read the knobs coming from the frontend configuration for blackhole detection (GLIDEIN_BLACKHOLE_NUMJOBS and GLIDEIN_BLACKHOLE_RATE)
 glidein_blackhole_numjobs=$(gconfig_get GLIDEIN_BLACKHOLE_NUMJOBS "$config_file")
 glidein_blackhole_rate=$(gconfig_get GLIDEIN_BLACKHOLE_RATE "$config_file")
+if [[ -n "$GLIDEIN_BLACKHOLE_RATE" && "$GLIDEIN_BLACKHOLE_RATE" =~ ^0(\.0*)?$ ]]; then
```
namrathaurs, line 77: Two questions here:
1. Why are we not checking both `glidein_blackhole_numjobs` and `glidein_blackhole_rate` variables here?

2. What does the regex pattern `^0(\.0*)?$` match? Is it any real number?

doc/factory/custom_vars.html
```diff
@@ -1709,13 +1709,14 @@ <h3><a name="frontend_vars"></a>Frontend Client Variables</h3>
                 This attribute represents number of jobs completed too quickly
                 before raising the black hole detection flag. More details in
                 black hole detection section, on the bottom of who webpage
-                (Lifetime of a Glidein).
+                (Lifetime of a Glidein). When defined, must be an integer bigger
+                than 0.
```
namrathaurs, line 1713: Is this in a new line rather than being a continuation of the previous line?

creation/web_base/condor_startup.sh
namrathaurs, line 1: A general question: upon looking at the proposed solution, I wondered why we would have to explicitly check for the presence of the two frontend variables in this case? 

In the past, I have worked mostly on adding new configuration knobs to the factory and from what I remember, there was no explicit checks to determine whether a particular configuration knob exists in the factory. Does this behavior not apply to the frontend as it is designed differently when compared to the factory?

namrathaurs (commented):
Proposed solution looks good, but I do have a few questions that I have added in this PR as comments

creation/web_base/condor_startup.sh
namrathaurs: The START expression is dependent on the `BLACKHOLE_CONDITION` expression as it is one of the many expressions. So if that ends up being malformed, then the whole glidein fails, which is what was observed.

creation/web_base/condor_startup.sh
mambelli: fixed tabs

doc/factory/custom_vars.html
mambelli: Same for html

creation/web_base/condor_startup.sh
mambelli: discussed

creation/web_base/condor_startup.sh
namrathaurs: 1. Only considering `glidein_blackhole_rate` since we have defined the following, as per our documentation:
```
Rate at which a Glidein is flagged as a black hole and retired. ... When undefined or set to 0.0, the black hole mechanism is disabled. See the blackhole detection section below ...
```

2. The regex pattern only matches `0`, `0.` `0.0`, `0.00...` (all of which means zero)

namrathaurs (approved):
Changes are ready to be merged!

PR #426, opened 2024-07-19
docstrings

uziahmed3:


lib/classadSupport.py
mambelli, line 1: This file contains some docstrings in different formats, e.g. in ClassadAdvertiser. Could they be converted to the Google format? And functions like `exe_condor_advertise` have incomplete docstrings (missing parameters). Could those be completed?

uziahmed3: Will work on that.

PR #428, opened 2024-08-02
Replacing xmlrunner with unittest-xml-reporting

mambelli:
Replacing xmlrunner with unittest-xml-reporting that works also in Py3.12
pip install of xmlrunner was giving an error


namrathaurs (approved):
Ready for merge!

PR #433, opened 2024-08-09
Moving FS authentication first to fix root job management

mambelli:
Moving FS authentication first otherwise root cannot remove/manage other users' jobs


namrathaurs (approved):
Changes are merge-ready!

PR #457, opened 2024-11-16
Fixed failed log rotation due to wrong log file creation time recording

### LINKED ISSUE #451: Logs do not rotate when they hit the max Mb limit
**Describe the bug**
Since a long time CMS has been experiencing issues with log rotation and disk filling up. Basically logs in `log/group_*` keeps growing until frontends just fills up the disk.

**To Reproduce**
Reduce the `max_mbytes="10.0"` param to a small number and verify logs are not rotated.

-   Priority: high
-   Stakeholders: CMS
-   Components: frontend

mambelli:
Fixed failed log rotation due to wrong log file creation time recording
POSIX provides only access, modification and metadata change times. Timed rotation should be based on creation time, available sometimes as `os.stat(path).st_birthtime`.
This fix uses the correct creation time and disables timed rotation when the creation time is not available.

This fixes #451


namrathaurs (approved):

PR #466, opened 2024-12-11
Always send SIGQUIT to HTCondor

mambelli:
Before SIGTERM was used for graceful shutdown. Now SIGQUIT is forwarded to HTCondor no matter the signal received by the Glidein (INT, TERM, QUIT)

This was requested by FactoryOps -J.Dost- under HTCondor -J.Frey- suggestion to reduce the leftover files problem when Glideins are killed


namrathaurs (approved):

PR #496, opened 2025-03-12
Support GPUs in the mapping of OSG_autoconf VOs

mmascher:


mambelli:
@mmascher The unit test is failing

mambelli (approved):

Issue #247, opened 2022-12-01
When updating GWMS on an existing system, ownership of files in `/var/lib/gwms-frontend/` are overridden 


vitodb:
**Describe the bug**
When updating GWMS on an existing system, ownership of files in `/var/lib/gwms-frontend/` are overridden.
This is affecting DecisionEngine setup where it is needed to have `decisionengine` user owning content of `/var/lib/gwms-frontend/`

**To Reproduce**
Install `glideinwms-vofrontend-glidein`, change ownership for files/folders in `/var/lib/gwms-frontend/`.
Update `glideinwms-vofrontend-glidein`, ownership of files/folders in `/var/lib/gwms-frontend/` is overridden to `root:glidein`

**Expected behavior**
Preserve ownership of files/folders in `/var/lib/gwms-frontend/`

**Screenshots**
If applicable, add screenshots and/or console outputs to help explain your problem.

**Info (please complete the following information):**

-   GlideinWMS version: 3.9.6-1.osg36.el8
-   Python version: 3.9.13
-   OS version: CentOS Stream 8
-   HTCondor version: 9.12.0
-   Priority: medium
-   Stakeholder:  DecisionEngine
-   Component: frontend

**Additional context**

PR #504, opened 2025-03-21
Make frontend library more friendly to other clients

mambelli:
Make the frontend library more friendly to other clients, e.g. the decision engine
- Moved the tokens and token passwords to the user home directory
- Made some files owned by the glidein group


BrunoCoimbra (approved):

mambelli:
This addressed partially issue #247

PR #532, opened 2025-05-28
Add the possibility to query all schedulers in a pool without listing…

### LINKED ISSUE #510: Add the possibility to query all schedulers in a pool without listing them
During one of the SI meetings we discussed that maintaining the list of schedds is an operational burden, especially during transitions. I mentioned the possibility to ask for a feature request and allow to specify schedd=ALL.

By specifying schedd=ALL GlideinWMS will query all the available schedulers in the pool:
`<schedd DN="ALL" fullname="ALL"/>`

-   Priority: medium
-   Stakeholders: CMS
-   Components: frontend

**Additional context**
During the glideinWMS meeting we discussed possible behaviors: the list of schedd is retrieved at frontend startup and saved in memory VS the list is retrieved at every cycle with a cache for (e.g.) one hour in case of repeated failures.

mmascher:
… them. Fixes #510

### DIFF
--- a/frontend/glideinFrontendElement.py
+++ b/frontend/glideinFrontendElement.py
@@ -165,6 +165,8 @@ class glideinFrontendElement:
         self.removal_requests_tracking = self.elementDescript.element_data["RemovalRequestsTracking"]
         self.removal_margin = int(self.elementDescript.element_data["RemovalMargin"])
 
+        self.schedd_cache = {"timestamp": None, "data": None}
+
         # Default behavior: Use factory proxies unless configure overrides it
         self.x509_proxy_plugin = None
 
@@ -394,7 +396,7 @@ class glideinFrontendElement:
 
         ## schedd
         idx = 0
-        for schedd_name in self.elementDescript.merged_data["JobSchedds"]:
+        for schedd_name in self.getScheddList():
             idx += 1
             forkm_obj.add_fork(("schedd", idx), self.get_condor_q, schedd_name)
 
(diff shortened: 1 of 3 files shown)


frontend/glideinFrontendElement.py
```diff
@@ -955,6 +957,38 @@ def iterate_one(self):
...
+        # logSupport.log.info("Getting list of schedulers from collector since 'ALL' has been used") # Too verbose. Keeping it around until production deployment
+        # If the admin specified ALL then query the condor collector to get the list of schedulers
+        res = glideinFrontendLib.getCondorStatusSchedds([None], constraint=None, format_list=[])
+
+        if res and None in res:
+            schedds = list(res[None].fetchStored().keys())
+
+            # Cache the result with timestamp
+            self.schedd_cache["data"] = schedds
+            self.schedd_cache["timestamp"] = time.time()
```
mambelli, line 977: This is more a comment than a change request.
Seems more a backup value than a cache. It is used only if the query failed and the timestamp is not used. Or is it used somewhere else I'm missing?

frontend/glideinFrontendElement.py
```diff
@@ -955,6 +957,38 @@ def iterate_one(self):
 
         return
 
+    def getScheddList(self):
+        # Get the original list from config
+        schedd_list = self.elementDescript.merged_data.get("JobSchedds", [])
+
+        if schedd_list != ["ALL"]:
+            # If the frontend admin specified a scheduler list use it.
```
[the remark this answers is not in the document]
mmascher: Thanks, Marco — good catch! Indeed, I only tested the case of sched=ALL in the global schedd configuration. In CMS, we don't use schedulers in groups.

I think option 2 covers the use case where the operator sets global=ALL, but some groups may specify a list of schedulers instead — perhaps for testing — and not use ALL.

I can’t think of a real use case for option 1, but maybe I’m just lacking imagination.

That said, I’m leaning toward validating the config such that if "ALL" is detected, we ensure it’s the only item in the list. If additional use cases arise, we can revisit this based on stakeholder feedback.

frontend/glideinFrontendElement.py
mmascher: Argh, I messed things up and I forgot to include an if :(

mmascher: Added the missing code to adderess cache usage here: https://github.com/glideinWMS/glideinwms/pull/532/commits/2e2154388197a5fbfb7b749b9e8f07b835e4aa7b

frontend/glideinFrontendElement.py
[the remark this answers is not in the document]
mmascher: @mambelli https://github.com/glideinWMS/glideinwms/pull/532/commits/9b96c8ee48cbd6708f4d61a586563a4624292baf contains the code to validate the schedd values.

mambelli (approved):

Issue #533, opened 2025-05-29
Add support for GLIDEIN_OVERLOAD_ENABLED to control CPU and memory overload


mmascher:
**Is your feature request related to a problem? Please describe.**

Currently, there is no way to conditionally enable or disable resource overloading (via `GLIDEIN_OVERLOAD_CPUS` and `GLIDEIN_OVERLOAD_MEMORY`) in a dynamic or controlled way. As soon as these variables are defined, they are always applied. This limits our ability to safely test or gradually roll out overload behavior across the fleet of pilots. We are forced to either remove these settings entirely or apply them universally, which increases operational risk. We are also interested in keeping a percentage of pilots non overloaded as a base for studies.

**Describe the solution you'd like**
Introduce a new configuration variable: `GLIDEIN_OVERLOAD_ENABLED`. This variable acts as a global switch to control whether overload values should be applied.

* If set to `"true"` (case-insensitive), and if `GLIDEIN_OVERLOAD_CPUS` or `GLIDEIN_OVERLOAD_MEMORY` are defined, overload will be applied.
* If set to `"false"` (default), overload is disabled regardless of the other settings.
* If set to a percentage string like `"80%"`, overload will be enabled with that probability, allowing for gradual or randomized rollout.

Example:

```xml
<attr name="GLIDEIN_OVERLOAD_ENABLED" const="True" glidein_publish="True" job_publish="True" parameter="True" publish="True" type="expr" value="80%"/>

```

This would enable overload logic for \~80% of pilots.

**Info (please complete the following information):**

- Stakeholders: CMS, FactoryOps
- Components: glidein, factory, documentation

PR #536, opened 2025-06-11
Add GLIDEIN_OVERLOAD_ENABLED to control CPU and memory overload. Fixe…

mmascher:
…s #533


mmascher:
Discussed in the meeting: the `setup_overload` function is duplicated in the two files. It would be nice to put the shared code somewhere. @mambelli is there an appropriate place already? Or should I create a new `setup_overload.sh` file?

mambelli (approved):

PR #571, opened 2025-08-02
Oda

Space-Paragon1:
Added and updated Google Docstrings for glideinFrontendConfig.py


mambelli (changes requested):
This PR is strange. It marks differences for over 300 files. Something is wrong. And it is not passing the pre-commit checks

mambelli:
Superseeded by #574

PR #572, opened 2025-08-02
Added Docstrings to glideinFrontendConfig.py

Space-Paragon1:


mambelli (changes requested):
This PR is strange. It marks differences for over 300 files. Something is wrong. And it is not passing the pre-commit checks

PR #591, opened 2025-09-11
add staticextra resource type to condor_startup.sh

### LINKED ISSUE #590: Add support for staticextra resource type to improve ioslot scheduling
The CMS Submission Infrastructure group was trying to add a dedicated *ioslot*.
Glideins were observed to have insufficient memory to schedule the desired number of jobs, leaving available CPUs idle. This mismatch between CPU and memory allocation prevented efficient resource utilization.

Introduce a new resource type `staticextra`, similar to `static`, that:

* Adds memory to the main partitionable slot instead of subtracting it.

This ensures that dedicated ioslots can be created without reducing the memory available to the main slot, resolving the CPU underutilization issue.

* Manually tuning `GLIDEIN_CPUS` and `GLIDEIN_MaxMemMBs` to balance CPU/memory, but this is error-prone and hard to maintain.

* **Priority:** high
* **Stakeholders:** CMS
* **Components:** glidein

mmascher:
Fixes #590

Introduce a new resource type `staticextra`, similar to `static`, that creates one dedicated slot per instance. Unlike `static`, the memory is not subtracted from the main partitionable slot but instead added to it.

Implementation details:

* Added `EXTRA_MEMORY_MB` knob and updated `MEMORY` expression.
* When `staticextra` is configured, memory for the slot is accumulated into `EXTRA_MEMORY_MB`.
* Keeps CPU handling unchanged (one vCPU per slot).

This allows deployments to dedicate resources with extra memory while preserving the full main slot capacity.

### DIFF
--- a/creation/web_base/condor_startup.sh
+++ b/creation/web_base/condor_startup.sh
@@ -826,8 +826,10 @@ EOF
 NEW_RESOURCES_LIST =
 EXTRA_SLOTS_NUM = 0
 EXTRA_CPUS_NUM = 0
+EXTRA_MEMORY_MB = 0
 EXTRA_SLOTS_START = True
 NUM_CPUS = \$(GLIDEIN_CPUS)+\$(EXTRA_SLOTS_NUM)+\$(EXTRA_CPUS_NUM)
+MEMORY = \$(GLIDEIN_MaxMemMBs)+\$(EXTRA_MEMORY_MB)
 
 # Slot 1 definition done before (fixed/partitionable)
 #SLOT_TYPE_1_PARTITIONABLE = FALSE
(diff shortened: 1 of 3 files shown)


creation/web_base/condor_startup.sh
```diff
@@ -945,6 +949,9 @@ SLOT_TYPE_${slott_ctr}_PARTITIONABLE = TRUE
 NUM_SLOTS_TYPE_${slott_ctr} = 1
 EOF
                     else
+                        if [[ "$res_opt" == "staticextra" ]]; then
+                            echo "EXTRA_MEMORY_MB = \$(EXTRA_MEMORY_MB)+${res_ram}" >> "$CONDOR_CONFIG"
```
mambelli, line 953: Condor is tricky with changes to the same variable. I'd recommend:
- After 822 add a line: `cc_resource_slots_mem_add=0` and here increment that variable `((cc_resource_slots_mem_add=cc_resource_slots_mem_add+res_ram))`
- Finally, after 970:
```
# Epilogue RAM
if [[ "$cc_resource_slots_mem_add" -ne 0 ]]; then
  echo "EXTRA_MEMORY_MB = ${cc_resource_slots_mem_add}" >> "$CONDOR_CONFIG"
fi
```

mambelli (changes requested):

mambelli (approved):

PR #600, opened 2025-10-13
Recognize EL/CentOS 10 worker nodes to select the correct HTCondor tarball

mambelli:
Added string parsing for EL 10 worker nodes
Added also strings for EL 11, Debian 14, and Ubuntu 26, just in case


namrathaurs (approved):
Changes ready to be merged! I did not see an update in CHANGELOG asoociated with this. Will it be added closer to the release?

PR #605, opened 2025-10-22
Remove coral frontend from first page animation

mmascher:


mambelli:
Let's replace the coral frontend with the Decision Engine instead.
Gimp source files are on the website:
https://glideinwms.fnal.gov/original_animations/simple_animation.xcf
Other files in the same folder: header2.xcf  header3.xcf  simple_animation_broken.xcf  simple_animation_op.xcf  simple_animation.xcf  simple_diagram.xcf

mambelli (changes requested):
Let's replace with the Decision Engine instead

PR #608, opened 2025-10-30
Added support also for htcondor v2 Python binding

mambelli:
Added support also for htcondor v2 Python binding, htcondor2 and classad2


namrathaurs (commented):
Only one minor comment.

PR #617, opened 2025-11-17
Use native systemd (systemctl command) in spec file

mambelli:
Switched spec file to use native systemd (systemctl) instead of sysVinit compatibility (service).

Previously, GWMS required the SystemV Init compatibility RPMs because of its use of the service command during the RPM installation.

### DIFF
--- a/doc/tags.yaml
+++ b/doc/tags.yaml
@@ -674,3 +674,22 @@ v3_10_16:
   NOTE:
     - Added httpd configuration to enhance security by disabling version headers and trace (PR#578)
     - Since 3.10.15, all job wrapper scripts must be "POSIX" compatible to be able to run on small images with Busybox. Added a relaxation in 3.10.16, setting `set +o posix` when Bash is available (PR#587)
+
+v3_10_17:
+  Date: November 13, 2025
+  Series: Stable
+  Tarball: false
+  Feature:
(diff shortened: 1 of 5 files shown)


doc/tags.yaml
```diff
@@ -674,3 +674,20 @@ v3_10_16:
...
     - Added httpd configuration to enhance security by disabling version headers and trace (PR#578)
     - Since 3.10.15, all job wrapper scripts must be "POSIX" compatible to be able to run on small images with Busybox. Added a relaxation in 3.10.16, setting `set +o posix` when Bash is available (PR#587)
+
+v3_10_17:
+  Date: November 13, 2025
+  Series: Stable
+  Tarball: false
+  Feature:
+    - Recognize EL/CentOS 10 worker nodes to select the correct HTCondor tarball (PR#600)
+    - Factory monitoring now showing Client Requested Idle Glideins; only keeping track of Factory adjusted Idle (PR##606, Issue#520)
```
namrathaurs: extra `#` in `... PR##606 ...`

namrathaurs (commented):
Comments need to be addressed before merging.

mambelli:
Pushed changes, adding also PR 618 preventively. Do not worry about tags.txt and history.html consistency; those are generated during the release process.

namrathaurs (approved):

PR #618, opened 2025-11-17
Fixed wrapper function check and removed exec from default_singularity_wrapper.sh

mambelli:
The wrapper check function flow was incorrect. 
Fixed also default_singularity_wrapper.sh not to exec and moved exec error notifications to the wrapper tail added in condor_startup.sh


namrathaurs (approved):

PR #631, opened 2025-12-23
Add full cvmfsexec support in GWMS

namrathaurs:
This PR supersedes #372 and contains the changes from the same. 

During the implementation and testing of full cvmfsexec support (coming through this PR), a local branch by the name `branch_v3_11_revised_latest` was in use. Since there were a couple of development releases in the meantime, this PR had to be rebased with `branch_v3_11` which resulted in feature-related changes to go out of sync. To avoid further confusion, a new branch was created based on `branch_v3_11_revised_latest` and was named `v311_revised_latest_dupe`, which is the source for this PR.

Dependencies:
- [x] #683


creation/create_cvmfsexec_distros.sh
```diff
@@ -75,7 +75,10 @@ build_cvmfsexec_distros() {
 	# check if the temp directory for cvmfsexec exists
 	ensure_directory_exists "$cvmfsexec_temp"
 
-	git clone $CVMFSEXEC_REPO "$cvmfsexec_latest" &> /dev/null
+	# download the cvmfsexec repository contents as a compressed tarball and extract its contents
+	# before extracting contents, make sure the `latest` directory for cvmfsexec exists
+	ensure_directory_exists "$cvmfsexec_latest"
+	curl -sL https://github.com/cvmfs/cvmfsexec/archive/master.tar.gz | tar -xz -C "$cvmfsexec_latest" --strip-components=1
```
shreyb: Similar to how line 78 had a reference to `$CVMFSEXEC_REPO`, it might make this more readable to set a global variable to this URL, and then reference the variable here.

shreyb (commented):
This mostly looks good to me.  I had just one comment.

creation/create_cvmfsexec_distros.sh
namrathaurs: Point taken, thanks for the suggestion @shreyb!

namrathaurs:
Hi Shreyas,

Somehow I could not see this comment of yours on Github, so replying here:

In Bash, the ,, is used to convert the value to lowercase. This helps resolve all unintentional specification of values of `glidein_cvmfs_require` (e.g. Required, REQUIRED, etc.) to required.
Possible values of `GLIDEIN_CVMFS_REQUIRE` (which is a factory config knob) include REQUIRED, PREFERRED, and NEVER.


Hope that helps,
Namratha

________________________________
From: Shreyas Bhat ***@***.***>
Sent: Wednesday, March 11, 2026 9:19 AM
To: glideinWMS/glideinwms ***@***.***>
Cc: Urs, NAMRATHA ***@***.***>; Author ***@***.***>
Subject: [EXT] Re: [glideinWMS/glideinwms] Add full cvmfsexec support in GWMS (PR #631)


@shreyb commented on this pull request.

________________________________

In creation/web_base/glidein_startup_ff.sh<https://github.com/glideinWMS/glideinwms/pull/631#discussion_r2918682476>:

>      mount_cvmfs_repos $gwms_cvmfsexec_mode $cvmfs_config_repo $cvmfs_add_repos
+    if [[ $? -ne 0 ]]; then
+        glidein_cvmfs_require=$(gconfig_get GLIDEIN_CVMFS_REQUIRE "$glidein_config")
+        glidein_cvmfs_require=${glidein_cvmfs_require,,}


What does the ,, do with the parameter?

creation/web_base/cvmfsexec_platform_select.sh
```diff
@@ -20,13 +20,15 @@ add_config_line_source=$(grep -m1 '^ADD_CONFIG_LINE_SOURCE ' "$glidein_config" |
 error_gen=$(gconfig_get ERROR_GEN_PATH "$glidein_config")
 
 # this script should run only when on-demand CVMFS is requested by the user/job,
-# so use the GLIDEIN_USE_CVMFSEXEC from the config file as a flag to determine
+# so use the GLIDEIN_USE_CVMFS from the config file as a flag to determine
```
mambelli, line 23: The documentation should be updated.
https://glideinwms.fnal.gov/doc.prd/factory/custom_vars.html#cvmfs_vars should include all CVMFS variables with current names.
The table should be updated: 
- the "cvmfsexec mode used" column should say "No" if CVMFS is not used/mounted (not N/A) and "Error" is the Glidein is failing (e.g. required and unable to provide it.
- Maybe 0/REQUIRED should still mount CVMFS
- What happens with 1/PREFERRED|OPTIONAL and no FUSE and no unpriv userns? That line seems missing

Maybe separate the table in 2: one to define the level of requirement for CVMFS, one for the actions depending on the value of that level

creation/web_base/cvmfsexec_platform_select.sh
```diff
@@ -20,13 +20,15 @@ add_config_line_source=$(grep -m1 '^ADD_CONFIG_LINE_SOURCE ' "$glidein_config" |
...
 # whether the rest of this script should be run or not
-use_cvmfsexec=$(gconfig_get GLIDEIN_USE_CVMFSEXEC "$glidein_config")
+use_cvmfs=$(gconfig_get GLIDEIN_USE_CVMFS "$glidein_config")
 # TODO: int or string?? if string, make the attribute value case insensitive
-#use_cvmfsexec=${use_cvmfsexec,,}
+#use_cvmfs=${use_cvmfs,,}
 
-if [[ "$use_cvmfsexec" -ne 1 ]]; then
+cvmfs_require=$(gconfig_get GLIDEIN_CVMFS_REQUIRE "$glidein_config")
```
mambelli, line 29: GLIDEIN_CVMFS_REQUIRE is not in the documentation. Is it the new name of GLIDEIN_CVMFS?

creation/web_base/glidein_startup_ff.sh
```diff
@@ -2142,8 +2142,20 @@ if [[ -n "$gwms_cvmfs_reexec" && "$gwms_cvmfs_reexec" == "yes" ]]; then
...
+    . "$work_dir"/cvmfs_helper_funcs_ff.sh
     mount_cvmfs_repos $gwms_cvmfsexec_mode $cvmfs_config_repo $cvmfs_add_repos
+    if [[ $? -ne 0 ]]; then
+        glidein_cvmfs_require=$(gconfig_get GLIDEIN_CVMFS_REQUIRE "$glidein_config")
+        glidein_cvmfs_require=${glidein_cvmfs_require,,}
+        # if exit status is non-zero, i.e something went wrong during mounting
+        if [[ "${glidein_cvmfs_require}" == "required" ]]; then
+            # if mount CVMFS is not successful, report an error and exit with failure exit code
+            logerror "Unable to mount CVMFS on worker node; aborting glidein setup (CVMFS ${glidein_cvmfs_require})"
+            exit 1
```
mambelli, line 2154: Should this check and failure be in one of the custom scripts?
Are you sure you can exit without invoking any cleanup function?

creation/create_cvmfsexec_distros.sh
```diff
@@ -6,13 +6,58 @@
...
+    curl -sL "$CVMFSEXEC_ARCHIVE" | tar -xz -C $temp_loc --strip-components=1
+    supported_types=$("$temp_loc"/makedist -m xxx 2>&1 | grep -v "not supported" | tail -n +3)
+    echo "$supported_types"     # printing the supported platforms...
+
+    # after listing the supported types, clean up the temporary location that was created
+    if [[ -n "$temp_loc" && -d "$temp_loc" ]]; then
+        rm -rf "$temp_loc"
+        return 0
+    else
+        error_handler "Something went wrong while cleaning up!"
```
mambelli, line 29: Usually functions are defined before using them (no shell error though).
Then you use error_handler that prints usage but I don't think this is an use error. And you have a return after the exit in the function

creation/create_cvmfsexec_distros.sh
```diff
@@ -6,13 +6,58 @@
 # This script generates cvmfsexec distributions for various cvmfs configurations
 # and supported machine types, as supported by the open-source cvmfsexec utility.
 
+## \brief Get all the machine types supported by cvmfsexec utility.
+## \param No parameters.
+## \returnval 0 if the supported machine types were obtained and cleanup of the temporary directory was successful, 1 if the supported machine types were obtained but cleanup of the temporary directory failed.
+# Check with Marco: is the return value necessary? since I'm not using the return value anywhere to determine the next course of action
```
mambelli, line 12: If functions do a task w/o printing error messages or exiting, then the return value is used to tell the caller if there was an error. You need it if it's checked by the caller, you can have it anyway for future uses or clarity in the function. In this case you do a mix of things depending on the path in the function. There is an exit, echo a return and a call to error_handler that does exit and return.

creation/create_cvmfsexec_distros.sh
```diff
@@ -6,13 +6,58 @@
...
+    # first, get the machine types supported by the cvmfsexec utility
+    machine_types=$(get_supported_machine_types)
+    # then, process the output to have a comma-separated list of machine types
+    # converting the multi-line output to single line string
+    machine_types=${machine_types//$'\n'/ }
+    # trimming leading/trailing spaces in the single line string
+    machine_types=$(echo "$machine_types" | awk '$1=$1')
+    # replacing spaces with commas
+    machine_types=${machine_types// /,}
+    echo "$machine_types"
```
mambelli, line 47: Can multiple machine types also be printed on the same line space separated by `get_supported_machine_types`?
Is `get_supported_machine_types` adding extra spaces?
If not to both, then something like this may be more efficient:
```
machine_types=$(get_supported_machine_types)
machine_types=${machine_types//$'\n'/,}
echo "${machine_types%,}"
```
The last one, `${%}` is likely not needed because $() is trimming the las newline, but just in case

creation/create_cvmfsexec_distros.sh
```diff
@@ -6,13 +6,58 @@
...
+}
+
 # Hardcoded variables
-CVMFSEXEC_REPO="https://www.github.com/cvmfs/cvmfsexec.git"
+# using the recommended URL format for legacy/custom repositories with master branch
+CVMFSEXEC_ARCHIVE="https://github.com/cvmfs/cvmfsexec/archive/master.tar.gz"
 DEFAULT_WORK_DIR="/var/lib/gwms-factory/work-dir"
-# TODO: periodically add rhel, suse and other derivatives as supported by cvmfsexec
+# TODO: periodically verify DEFAULT_MACHINE_TYPES to ensure rhel, suse and other derivatives as supported by cvmfsexec are included in the list
 # NOTE: Although rhel9-x86_64 is supported, el7 tools might not work with el9 files (as suggested by Dave Dykstra) as of July 03, 2023
```
mambelli, line 55: Maybe move these notes where you define set_default_machine_types

creation/create_cvmfsexec_distros.sh
```diff
@@ -29,10 +74,13 @@ or a comma-separated list of values from the options {osg|egi|default}.
 PLATFORMS_LIST (optional): indicates machine types (platform- and architecture-based)
 for which distributions is to be built. Can be empty, a single value or a
 comma-separated list of values from the options {rhel9-x86_64|rhel8-x86_64|rhel7-x86_64|suse15-x86_64|rhel8-aarch64|rhel8-ppc64le}.
```
mambelli, line 76: Should the default from `$(set_default_machine_types)` be added?
Is DEFAULT_MACHINE_TYPES, the defaults, all available or both? If not the first, then the list above should be printed and something like "Defaults to all available platforms."

creation/create_cvmfsexec_distros.sh
```diff
@@ -213,7 +269,7 @@ elif [[ "$2" =~ $re_mtypes ]]; then
 	machine_types=$2
 else
 	# handle if there were typos in the arguments passed
-	error_handler "Invalid platform provided. Must be empty or comma-separated list with (rhel9-x86_64|rhel8-x86_64|rhel7-x86_64|suse15-x86_64|rhel8-aarch64|rhel8-ppc64le), not '$2'"
+	error_handler "Invalid platform provided. Must be empty or comma-separated list with (${DEFAULT_MACHINE_TYPES//,/|}), not '$2'"
```
mambelli, line 272: Should this be just a warning and then fail only if the machine type is not found in the current list? This would allow to run if DEFAULT_MACHINE_TYPES gets obsolete. Is DEFAULT_MACHINE_TYPES, the defaults, all available or both?

creation/create_cvmfsexec_distros.sh
```diff
@@ -200,8 +256,8 @@ fi
 
 # after confirming that the remaining number of arguments to be either 1 or 2
 re_sources="^((osg|egi|default),)*(osg|egi|default),?$"
-# TODO: update the regex for other machine types upon checking with `makedist -h` periodically
-re_mtypes="^((rhel(7|8|9)|suse15)(-(x86_64|aarch64|ppc64le),?))+$"
+# TODO: update the regex for other/newly supported machine types upon verifying DEFAULT_MACHINE_TYPES or, alternatively, `makedist -h`
+re_mtypes="^((rhel(7|8|9|10)|suse15)(-(x86_64|aarch64|ppc64le),?))+$"
```
mambelli, line 260: Maybe mention where DEFAULT_MACHINE_TYPES is defined to easily compare/update all lines that have to be kept consistent.

doc/factory/custom_vars.html
```diff
@@ -2892,10 +2891,9 @@ <h2><a name="cvmfs_vars"></a>CVMFS Variables</h2>
         </p>
         <p>
           The table below describes the attributes that can be configured for
-          use with the glidein to allow CVMFS to be provisioned on-demand. These
-          variables are configured at the glidein level so that it applies to
-          all the entries that are configured in the factory's configuration
-          file.
+          use with the Glidein to allow CVMFS to be provisioned on-demand. These
+          attributes are configured at the Glidein level so that it applies to
+          all the entries that are defined in the Factory's configuration file.
```
mambelli, line 2896: Are these only global or could be also defined in the entry?

doc/factory/custom_vars.html
```diff
@@ -2916,18 +2914,17 @@ <h2><a name="cvmfs_vars"></a>CVMFS Variables</h2>
...
+            <td>Frontend</td>
             <td>
               <p>
-                Not set by default. This variable controls whether
-                <i>cvmfsexec</i> utility is used to mount CVMFS. Possible values
-                are 0 or 1. A value of 0 indicates that glidein will not use
-                <i>cvmfsexec</i> to mount CVMFS on the worker noden. A value of
-                1 indicates <i>cvmfsexec</i> will be used to mount CVMFS on the
-                worker node by the glidein.
+                Not set by default. This attribute controls whether the user/job
```
mambelli, line 2923: Does not set equates to 0, 1 or a 3rd option?

doc/factory/custom_vars.html
```diff
@@ -2938,47 +2935,96 @@ <h2><a name="cvmfs_vars"></a>CVMFS Variables</h2>
             <td>Factory</td>
             <td>
               <p>
-                Not set by default. This variable controls the origin (source)
+                Not set by default. This attribute controls the origin (source)
```
mambelli, line 2938: Does not set equates to default or something different?

doc/factory/custom_vars.html
```diff
@@ -2938,47 +2935,96 @@ <h2><a name="cvmfs_vars"></a>CVMFS Variables</h2>
...
-                REQUIRED - if unable to mount, report an error and abort
-                mounting);<br />
-                PREFERRED - if unable to mount, notify and continue normally;<br />
-                OPTIONAL - continue if unable to mount (similar to
-                PREFERRED);<br />
-                NEVER - only exit if mounting fails<br />
+                Not set by default. This attribute, in conjunction with
+                GLIDEIN_USE_CVMFS (defined in the Frontend), specifies the
+                site/factory policy to be used when provisioning CVMFS. This
+                also helps in determining the behavior of the Glidein behavior
```
mambelli, line 2957: the behavior of the Glidein ~~behavior~~

doc/factory/custom_vars.html
```diff
@@ -2938,47 +2935,96 @@ <h2><a name="cvmfs_vars"></a>CVMFS Variables</h2>
...
-                Not set by default. This variable controls the error handling
-                behavior of the glidein in case of failures during the mounting
-                of CVMFS. Possible values are:<br />
-                REQUIRED - if unable to mount, report an error and abort
-                mounting);<br />
-                PREFERRED - if unable to mount, notify and continue normally;<br />
-                OPTIONAL - continue if unable to mount (similar to
-                PREFERRED);<br />
-                NEVER - only exit if mounting fails<br />
+                Not set by default. This attribute, in conjunction with
```
mambelli, line 2954: Does not set equates to one of the possible values? 
Then would be nice to add also a link to the table of behaviots in combination w/ GLIDEIN_USE_CVMFS

test/bats/creation_web_base_cvmfs_helper_funcs.bats
```diff
@@ -0,0 +1,367 @@
...
+    export -f arch
+    export -f uname
+    export -f unshare
+    export -f dnf
+    export -f fusermount3
+    export -f getent
+    export -f cat
+
+    run perform_system_check
+    run log_all_system_info
```
mambelli, line 355: I don't understand where you test the outputs/behaviors.
Is the test just to see if the functions run or fail?

mambelli (changes requested):

creation/web_base/cvmfsexec_platform_select.sh
namrathaurs: The documentation link in the comment above (https://glideinwms.fnal.gov/doc.prd/factory/custom_vars.html#cvmfs_vars) is the one for production version. With respect to production version of on-demand CVMFS provisioning (via mode 1), the names of CVMFS variables has NOT CHANGED.

The documentation for development series (https://glideinwms.fnal.gov/doc.dev/factory/custom_vars.html#cvmfs_vars) is being updated via this PR to update the newer names for CVMFS variables. See `#cvmfs_vars` section in `doc/factory/custom_vars.html` file of this PR. If you'd like to preview the updated documentation page (from this PR), you can access it [here](https://htmlpreview.github.io/?https://github.com/namrathaurs/glideinwms/blob/v311_revised_latest_dupe/doc/factory/custom_vars.html#cvmfs_vars).

Some notes:
* Two of the three CVMFS variables have newer names:
    1. `GLIDEIN_USE_CVMFSEXEC` is changing to `GLIDEIN_USE_CVMFS`.
    2. `GLIDEIN_CVMFS` is changing to `GLIDEIN_CVMFS_REQUIRE`.
    3. `CVMFS_SRC` remains unchanged.
* `GLIDEIN_CVMFS_REQUIRE` can only take one of three values: `REQUIRED`, `PREFERRED` or `NEVER` (as opposed to four values which includes the three plus `OPTIONAL` values). Decided to remove `OPTIONAL` from the new list of possible values since both `PREFERRED` and `OPTIONAL` might end up having the same behavior and found the following in current (production) documentation as well:
    
    > REQUIRED - if unable to mount, report an error and abort mounting);
    > **PREFERRED - if unable to mount, notify and continue normally;
    > OPTIONAL - continue if unable to mount (similar to PREFERRED);**
    > NEVER - only exit if mounting fails

<br/>

> Maybe separate the table in 2: one to define the level of requirement for CVMFS, one for the actions depending on the value of that level

I like the suggestion to separate the table into two separate tables and updated to reflect the same. There seems to be redundant information (in the new table and the table containing on-demand CVMFS knobs). I'm retaining the redundancy so that you can evaluate if your suggestion for two tables is still necessary since I already had the information put under the CVMFS knobs table.

creation/web_base/cvmfsexec_platform_select.sh
namrathaurs: Yes, `GLIDEIN_CVMFS_REQUIRE` is the new name of `GLIDEIN_CVMFS`. The new name will not be reflected in the documentation for production version because the production version uses old names. 

The new name has been updated in the documentation for development version, which is coming in as a change through this PR, and as can be seen [here](https://htmlpreview.github.io/?https://github.com/namrathaurs/glideinwms/blob/v311_revised_latest_dupe/doc/factory/custom_vars.html#cvmfs_vars).

creation/create_cvmfsexec_distros.sh
namrathaurs: > Can multiple machine types also be printed on the same line space separated by get_supported_machine_types?

`get_supported_machine_types` is a custom created function that is solely used to get all the possible machine types supported by cvmfsexec utility. By default (which is native cvmfsexec behavior), machine types are not printed on the same line, and hence causes the list of machine types returned to be a multi-line output string.
```
# output returned by native cvmfsexec
...
...
  rhel7-x86_64
  rhel8-aarch64 rhel8-ppc64le rhel8-x86_64
  rhel9-aarch64 rhel9-ppc64le rhel9-x86_64
  rhel10-x86_64
  suse15-aarch64 suse15-x86_64
...
...
```
<br/>

> Is get_supported_machine_types adding extra spaces?

No, the `get_supported_machine_types()` is not adding extra spaces, but the multi-line output returned by native cvmfsexec contains extra spaces (i.e. leading/trailing spaces). So, designed `set_default_machine_types()` to explicitly handle the multi-line output by removing leading/trailing spaces and separating the values using commas.
<br/>

>If not to both, then something like this may be more efficient:
>```
> machine_types=$(get_supported_machine_types)
> machine_types=${machine_types//$'\n'/,}
> echo "${machine_types%,}"
> ```
Tried the more efficient approach as you suggested but it still contains extra spaces and is still not a purely comma-separated list of values:
```
MYDEBUG (line 1): machine_types =   rhel7-x86_64
  rhel8-aarch64 rhel8-ppc64le rhel8-x86_64
  rhel9-aarch64 rhel9-ppc64le rhel9-x86_64
  rhel10-x86_64
  suse15-aarch64 suse15-x86_64
MYDEBUG (line2) : machine_types =   rhel7-x86_64,  rhel8-aarch64 rhel8-ppc64le rhel8-x86_64,  rhel9-aarch64 rhel9-ppc64le rhel9-x86_64,  rhel10-x86_64,  suse15-aarch64 suse15-x86_64
# output for line 3
  rhel7-x86_64,  rhel8-aarch64 rhel8-ppc64le rhel8-x86_64,  rhel9-aarch64 rhel9-ppc64le rhel9-x86_64,  rhel10-x86_64,  suse15-aarch64 suse15-x86_64
```

creation/create_cvmfsexec_distros.sh
namrathaurs: Good catch! Line 76 has been updated to dynamically use the value in `DEFAULT_MACHINE_TYPES` variable to keep the comma-separated list of values up to date. 

`DEFAULT_MACHINE_TYPES` always contains all available platforms and is used as the default value when `machine_types` parameter is empty/not specified (this parameter comes from the Factory config `cvmfsexec_distros` tag):
```
....
if [[ -z "$2" ]]; then
	machine_types=$DEFAULT_MACHINE_TYPES
elif [[ "$2" =~ $re_mtypes ]]; then
	machine_types=$2
else
	# handle if there were typos in the arguments passed
	error_handler "Invalid platform provided. Must be empty or comma-separated list with (${DEFAULT_MACHINE_TYPES//,/|}), not '$2'"
fi
....
```

creation/create_cvmfsexec_distros.sh
namrathaurs: I think I can remove the TODO note since `DEFAULT_MACHINE_TYPES` will always be supporting the same list of platforms as cvmfsexec utility. This is because `DEFAULT_MACHINE_TYPES` relies on `set_default_machine_types()` which internally relies on `get_supported_machine_types()` and this always returns the most up-to-date list of supported platforms.

I'm not sure if moving the NOTE item on line 55 to `set_default_machine_types()` would make sense. Maybe we could discuss this further.

creation/create_cvmfsexec_distros.sh
namrathaurs: Moved the TODO note from line 259 to where `DEFAULT_MACHINE_TYPES` variable is defined as suggested.

doc/factory/custom_vars.html
namrathaurs: fixed

creation/create_cvmfsexec_distros.sh
namrathaurs: `DEFAULT_MACHINE_TYPES` always contains all available platforms and is used as the default value when `machine_types` parameter is empty/not specified (this parameter comes from the Factory config `cvmfsexec_distros` tag when the `create_cvmfsexec_distros.sh` script is internally invoked during Factory reconfig).

doc/factory/custom_vars.html
namrathaurs: By "not set by default", I meant to convey that the Factory does not have the `GLIDEIN_USE_CVMFS` knob in the Factory config, which is considered that the Factory is not requesting on-demand CVMFS provisioning. Only if the Factory knob exists in the config along with a set value of 0 or 1, it is considered that the Factory is requesting for on-demand CVMFS provisioning.

test/bats/creation_web_base_cvmfs_helper_funcs.bats
namrathaurs: The latter; I believe I added this test to see if the functions `perform_system_check` and `log_all_system_info` run as expected in a pre-requisite full system check workflow. I do have added other tests in the same test script where I test the outputs/behaviors of the two functions that is in this test.

creation/create_cvmfsexec_distros.sh
namrathaurs: Great points here! As per your suggestions, I have moved the function definitions before using them and changed a couple of lines in the code to use simple `echo` statements rather than the `error_handler` function to align with the use of `return`.

creation/create_cvmfsexec_distros.sh
namrathaurs: While I was sure that I wanted to retain the return value for clarity and/or future use, wanted to double-check if it's okay to have the return statements in the code even though the caller does not need it.

There is no longer a mix of things as I have updated the use of `error_handler` function to instead use a simple `echo` to provide a logical path of print an error message and exit or print message and return with an appropriate value.

doc/factory/custom_vars.html
namrathaurs: By "not set by default", I meant to convey that the Factory does not have the `CVMFS_SRC` knob in the Factory config, which is when the Factory does not request on-demand CVMFS provisioning. Only when this knob exists in the Factory config along with one of the possible values, it is considered that the Factory will use this knob towards on-demand CVMFS provisioning.

doc/factory/custom_vars.html
namrathaurs: By "not set by default", I meant to convey that the Factory does not have the `GLIDEIN_CVMFS_REQUIRE` knob in the Factory config, which is when the Factory does not request on-demand CVMFS provisioning. Only when this knob exists in the Factory config along with one of the possible values, it is considered that the Factory will use this knob for on-demand CVMFS provisioning.

doc/factory/custom_vars.html
namrathaurs: I _think_ these can also be defined in the entry. During the last round of testing that was performed, I did not specifically test with the three knobs defined at the entry-level. Will test this and confirm the behavior.

creation/web_base/glidein_startup_ff.sh
namrathaurs: While I would have liked to include this check and failure to be in one of the custom scripts, I took another look at the `cvmfs_setup.sh` custom script and the `cvmfs_helper_funcs.sh` script to see if this can be made possible. It does not look like it might be possible to include in the custom script, but I can possibly try tweaking the `mount_cvmfs_repos` function in the helper script. Is that something reasonable to do instead?

Would like to further discuss with you regarding this to present my idea and get your feedback.

PR #648, opened 2026-02-26
Adding fallback mechanism for proxy availability test to avoid complete Glidein failure

### LINKED ISSUE #521: Requesting ability to configure glideins to continue without squid if it encounters squid errors
Currently to tell glideins to use a squid, we put `proxy_url="OSG"` into the entry config. However if this is set, and the glidein startup script fails when trying the squid declared at the site in OSG_SQUID_LOCATION env var, the glidien will throw a validation failure and terminate.

I would like an optional additional field to tell the glidein "continue without using squid on squid error" The current behavior of failing on validation error can stay the default if this flag is not set.

Jeff

namrathaurs:
Closes #521.

While the existing Glidein startup process included a test for proxy availability test, the Glidein would fail entirely if the proxy was found to be unavailable, resulting in a validation error. The requested improvement, as outlined in the associated issue, was to fall back to the `proxy_url="None"` behavior should the proxy availability test fails instead of failing the Glidein completely.

### DIFF
--- a/creation/web_base/glidein_startup.sh
+++ b/creation/web_base/glidein_startup.sh
@@ -834,6 +834,29 @@ add_config_file() {
     fi
 }
 
+test_proxy_availability() {
+    # Return 1 if the URL retrieval via HTTP proxy is failing,
+    # 0 otherwise (success or `wget` and `curl` not available)
+    local proxy_url_test=$1
+    local test_file=$2
+
+    # check whether wget or curl is available for use
+    if wget --version > /dev/null 2>&1; then
+        # wget found, use wget for the test
+        if ! wget -qO /dev/null -e use_proxy=yes -e http_proxy="$proxy_url_test" "${test_file}" 2>&1; then
+            return 1
+        fi
+    elif curl --version > /dev/null 2>&1; then
+        # curl found, use curl for the test
+        if ! curl -so /dev/null --proxy "$proxy_url_test" "$test_file" 2>&1; then
+            return 1
+        fi
+    else
+        echo "Neither wget nor curl command found. Skipping proxy availability test." 1>&2
+    fi
+    # implied return of exit status 0
+}
+
 #####################
 # Fetch a single file
 #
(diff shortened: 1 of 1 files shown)


creation/web_base/glidein_startup.sh
```diff
@@ -834,6 +834,35 @@ add_config_file() {
...
+        # found wget installed; use wget for the test
+        if ! wget -qO /dev/null -e use_proxy=yes -e http_proxy="$proxy_url_test" "${test_file}" 2>&1; then
+            echo -e "Configured proxy ($proxy_url_test) might be unavailable!\n" 1>&2
+            return 1
+        fi
+    fi
+
+    echo -e "Configured proxy ($proxy_url_test) available!\n" 1>&2
+    # implied return of exit status 0
+}
```
mambelli, line 858: I'd reduce the messages, stderr is more for errors/warnings than debug
if/elif may be more linear. I moved the messages in the caller
```
test_proxy_availability() {
    # Return 1 if the URL retrieval via HTTP proxy is failing, 
    # 0 otherwise (success or wget and curl not available)
    local proxy_url=$1
    local test_url=$2

    # check whether wget or curl is available for use
    if wget --version > /dev/null 2>&1; then
        # wget found; use wget for the test
        if ! wget -qO /dev/null -e use_proxy=yes -e http_proxy="$proxy_url" "$test_url" 2>&1; then
            return 1
        fi
    elif curl --version > /dev/null 2>&1; then
        # curl found; use curl for the test
        if ! curl -so /dev/null --proxy "$proxy_url" "$test_url" 2>&1; then
            return 1
        fi
    else
        echo "Neither wget nor curl command found. Skipping proxy availability test." 1>&2
    fi
    # implied return of exit status 0
}
```

creation/web_base/glidein_startup.sh
```diff
@@ -1479,6 +1508,13 @@ if [ "${proxy_url}" = "OSG" ]; then
     fi
 fi
 
+# testing proxy availability now...
+if [[ -n "${proxy_url}" && "${proxy_url}" != "None" ]]; then
+    if ! test_proxy_availability "${proxy_url}" "${repository_url}/glidein_startup.sh"; then
+        proxy_url="None"
+    fi
+fi
```
mambelli, line 1511: Print here the messages
```
if [[ -n "${proxy_url}" && "${proxy_url}" != "None" ]]; then
    if ! test_proxy_availability "${proxy_url}" "${repository_url}/glidein_startup.sh"; then
        echo "Resetting the HTTP proxy. The configured one ($proxy_url) seems unavailable!" 1>&2

        proxy_url="None"
# if you want the else, uncomment, otherwise you can omit if all goes OK. Stderr is more for warnings than debug
#    else
#        echo "Configured proxy ($proxy_url) available!" 1>&2
    fi
fi
```

mambelli (changes requested):
I'm sending a proposal that simplifies the function.
Feel free to take what you like.
I'd like at least to move the messages (echo) to the caller
Thanks

creation/web_base/glidein_startup.sh
namrathaurs: Going by the idea that stderr is more for warnings, it makes sense to skip the else block you suggested. Additionally, it might be better if we do not include the else block as that would result in false reporting of the message for the case when `wget` or `curl` is not available (this case also returns 0).

mambelli (approved):

PR #662, opened 2026-04-23
Fix incorrect Arch requirement for ARM pilots

### LINKED ISSUE #661: Incorrect Arch requirement (X86_64) set for ARM jobs due to GlideinWMS logic
## **Describe the bug**

When submitting jobs intended for ARM (`aarch64`) resources, the generated HTCondor job requirements incorrectly request `X86_64` instead. This results in jobs remaining idle despite available ARM resources at the site.

The issue originates from GlideinWMS logic that does not set any requirements. By default HTCondor adds to the requirements the default architecture of the submit host (x86_64).

---

## **To Reproduce**

1. Submit jobs targeting ARM resources (e.g., CERN CE `ce512.cern.ch`).
2. Inspect the job via:

   ```
   condor_q <jobid> -af:r Requirements
   ```
3. Observe that the job requirements contain:

   ```
   (TARGET.Arch == "X86_64") && ...
   ```

   instead of `aarch64`.

---

## **Expected behavior**

Jobs intended for ARM resources should explicitly request:

```
(TARGET.Arch == "aarch64")
```

and match available ARM nodes.

The architecture should not silently fall back to the submit host architecture when ARM is intended.

---

## **Info (please complete the following information):**

* Priority: high
* Stakeholders: CMS
* Components: factory

---

## **Additional context**

* CERN ARM resources are available are:
   * ~290 slots (`aarch64/AlmaLinux9`)
   * ~2320 total CPUs
* HTCondor defaults `Arch` and `OpSys` to the submit machine if not explicitly specified.

mmascher:
Ensure aarch64 is explicitly enforced when requested, avoiding defaulting to submit host architecture.

Fixes #661

### DIFF
--- a/creation/lib/cgWCreate.py
+++ b/creation/lib/cgWCreate.py
@@ -354,6 +354,13 @@ class GlideinSubmitDictFile(cgWDictFile.CondorJDLDictFile):
         # Add this because GLIDEIN_OVERLOAD_ENABLED might be set in the frontend
         self.add("job_ad_information_attrs ", "GlideinOverloadEnabled")
 
+        arch_attr = next((ga for ga in glidein_attrs if ga["name"].upper() == "GLIDEIN_ARCH"), None)
+        if arch_attr is not None and arch_attr["value"] == "aarch64":
+            self.add(
+                "Requirements",
+                '(TARGET.Arch == "aarch64") && (TARGET.OpSys == "LINUX") && (TARGET.Disk >= RequestDisk) && (TARGET.Memory >= RequestMemory)',
+            )
+
     def populate_submit_attrs(self, submit_attrs, gridtype, attr_prefix=""):
         for submit_attr in submit_attrs:
             if (


mambelli:
@mmascher Please rebase off the current master. This will fix the pre-commit error, not sure about the unit test. Retry? Seems transitory?
Then I have a comment to the code

creation/lib/cgWCreate.py
```diff
@@ -354,6 +354,13 @@ def populate_standard_grid(
         # Add this because GLIDEIN_OVERLOAD_ENABLED might be set in the frontend
         self.add("job_ad_information_attrs ", "GlideinOverloadEnabled")
 
+        arch_attr = next((ga for ga in glidein_attrs if ga["name"].upper() == "GLIDEIN_ARCH"), None)
+        if arch_attr is not None and arch_attr["value"] == "aarch64":
+            self.add(
+                "Requirements",
+                '(TARGET.Arch == "aarch64") && (TARGET.OpSys == "LINUX") && (TARGET.Disk >= RequestDisk) && (TARGET.Memory >= RequestMemory)',
+            )
```
mambelli, line 362: Why adding the requirement with also OpSys, disk, memory, ... only for the ARM architecture and not others?
Did I miss something?

mambelli (dismissed):

creation/lib/cgWCreate.py
mmascher: I basically copied and pasted the current `Requirement` for 86_64 and just changed just the architecture. It was a quick change to get the test going and not stall here.

But this is the part I am not convinced and I wanted to discuss. I am aware hardcoding the expression is bad design and if you have a better idea I am all ears!

mambelli (approved):

PR #680, opened 2026-06-22
Made check_proxy.sh more robust and verbose

### LINKED ISSUE #679: Error in scripts verifying x509
**Describe the bug**
The script `check_proxy.sh` verifies x509 proxies for the purposes of joining the User pool. If there is no proxy or the proxy is expired it prints a message and if GLIDEIN_CONDOR_TOKEN is not set (i.e. there is no IDTOKEN to join the pool) it fails.

Almost no site uses x509 any more. For some glideins GLIDEIN_CONDOR_TOKEN is not set and the script fails. 
This could be a problem with the generation of IDTOKENS, which should be automatic, or with the detection of the token and propagation of GLIDEIN_CONDOR_TOKEN.

Here is the frontend configuration:
https://github.com/opensciencegrid/tiger-osg-config/blob/master/manifests/production/vo-frontend-ucsd/frontend.xml

**To Reproduce**
Not reproduced in the test environment so far

**Expected behavior**
Glideins should not fail and use the IDTOKEN

**Screenshots**
This is the report from Jeff Dost

I'm trying to understand these errors that may be related in the UCSD CMS glidein pool. This FE / Collector are also on Tiger, they seem to still be running but we can't access them. What I can access is the glidein logs from the NRP factory. It looks like they are failing with:
```
 Validation failed in main/check_proxy.sh.

    Could not obtain -timeleft from grid-proxy-info/voms-proxy-info/openssl[12:07 PM]this is weird to me as I don't think we use x509 at all anymore in this pool
```
https://github.com/opensciencegrid/tiger-osg-config/blob/master/manifests/production/vo-frontend-ucsd/frontend.xml

More info from Diego:
so, the script is skipped if this var is set: GLIDEIN_CONDOR_TOKEN
and it is not set in the failing pilots. I'm guessing whatever is going wrong in tiger is preventing us from getting fresh IDTokens.

More from Jeff:
GLIDEIN_CONDOR_TOKEN wassn't missing in the factory ad but it was set to empty string. It was for all entries.
What i didn't check was if we were even getting the token forwarded from the FE to factory and if it made it to disk.. we already restarted the FE and things recovered

-   GlideinWMS version: Factory 3.11.4-1, Frontend 3.10.17-1
-   Python version:
-   OS version:
-   HTCondor version:
-   Priority: high
-   Stakeholders: CMS, FactoryOps, OSG

**Additional context**
This should be investigated, reproduced and fixed.

For the script both GLIDEIN_CONDOR_TOKEN missing or empty will cause the same error. The fact that was empty points me towards the FE non generating/forwarding the IDTOKENS. It should generate per-site tokens, using the key shared with the Pool collector. I would expect the FE complaining and not even submit requests if there are problems, but we should check. We'll try to check some failure modes like missing key (may fail at reconfig), removing the key, not forwarding it, ...

Long term. There are 2 scripts verifying and moving credentials: check_proxy.sh and setup_x509.sh. These should be combined in a single file setup_credentials.sh (git mv of setup_x509.sh) which is robust, has bats tests, works also with older x509 but is centered on tokens (refactored credentials, logging credentials, ...). At least for 3.11+

mambelli:


creation/web_base/check_proxy.sh
```diff
@@ -30,45 +30,53 @@ export GLIDEIN_CONDOR_TOKEN
 #if an IDTOKEN is available, continue.  Else exit
 exit_if_no_token() {
     if [ !  -e "$GLIDEIN_CONDOR_TOKEN" ]; then
+        echo "check_proxy.sh: idtoken '$GLIDEIN_CONDOR_TOKEN' not readable, failing"
```
namrathaurs: The `if` condition seems to be checking for existence of `$GLIDEIN_CONDOR_TOKEN` but the message says that the idtoken is not readable, which is not the same thing right?

creation/web_base/check_proxy.sh
```diff
@@ -30,45 +30,53 @@ export GLIDEIN_CONDOR_TOKEN
...
-
-    if [ "x$2" = "x" ]; then
-        epoch_now=$(date +%s)
-    else
-        epoch_now=$2
-    fi
-
+    # Extract dates
+    output=$(openssl x509 -noout -subject -dates -in "$cert_pathname" 2>/dev/null) || return 2
+    start_date=$(echo $output | sed 's/.*notBefore=\(.*\).*not.*/\1/g')  # intentional word splittig to remove newline
```
namrathaurs: typo: `splitting`

creation/web_base/check_proxy.sh
```diff
@@ -30,45 +30,53 @@ export GLIDEIN_CONDOR_TOKEN
...
+    end_date=$(echo $output | sed 's/.*notAfter=\(.*\)$/\1/g')  # intentional word splittig to remove newline
```
namrathaurs: typo: `splitting`

creation/web_base/check_proxy.sh
```diff
@@ -127,6 +138,5 @@ get_x509_expiration() {
 get_x509_expiration
 X509_EXPIRE="$RETVAL"
 
-"$error_gen" -ok "check_proxy.sh" "proxy" "$X509_USER_PROXY" "proxy_expire" "$(date --date=@"$X509_EXPIRE" +%Y-%m-%dT%H:%M:%S%:z)" "cert_dir" "$X509_CERT_DIR"
-
+"$error_gen" -ok "check_proxy.sh" "proxy" "$X509_USER_PROXY" "proxy_expire" "$(date --date=@"$X509_EXPIRE" +%Y-%m-%dT%H:%M:%S%:z)" "cert_dir" "$X509_CERT_DIR" "token" "$GLIDEIN_CONDOR_TOKEN"
```
namrathaurs, line 141: In `"$(date --date=@"$X509_EXPIRE" +%Y-%m-%dT%H:%M:%S%:z)"`, does having double quotes enclosed within `"$(...)"` be problematic?

namrathaurs (changes requested):
Minor comments need addressing.

creation/web_base/check_proxy.sh
mambelli: GLIDEIN_CONDOR_TOKEN is an idtoken, the one to join the pool

creation/web_base/check_proxy.sh
mambelli: fixed

creation/web_base/check_proxy.sh
mambelli: fixed

creation/web_base/check_proxy.sh
mambelli: no, it's ok

creation/web_base/check_proxy.sh
namrathaurs: Sorry I was not clear in my comment now that I read it again.

What I meant to say/ask was that the check seems to be for the existence of the idtoken but the message printed says “readable”. Is that intentional?

Otherwise, either the condition should change to if `[ ! -r "$GLIDEIN_CONDOR_TOKEN" ]` (instead of `-e`) or the message should read idtoken `'$GLIDEIN_CONDOR_TOKEN' does not exist, failing` (instead of `not readable`).

namrathaurs: The recent clarification was addressed and fixed.

namrathaurs (approved):
Comments addressed and changes are ready to be merged.

PR #687, opened 2026-07-13
Added a compatibility import for the type names changed in cryptography v40+

mambelli:
EL9 python3-cryptography RPM includes v36 that has only the uppercase names. The try-except will allow compatibility with both new and old syntax. This can be removed once only cryptography v40+ is used


namrathaurs (approved):

PR #688, opened 2026-07-21
Changing `run_with_timeout`'s default timeout

namrathaurs:
The `run_with_timeout` function was introduced as a watchdog timer recently as Jeff Dost (Factory Ops) and Mats Rynge (Frontend Ops) had encountered a situation where the execution of `cvmfs_helper_funcs.sh` was found hanging on a site (prior to the timeout function being added) which was resulting in all the Glideins to fail.

The idea behind the timeout function is as follows:
* if the command supplied to the `run_with_timeout function` takes long to run and does not complete within the set timeout, then the command execution is killed.
* if the command supplied to the `run_with_timeout function` does not take too long to run and it completes within the set timeout, then the program execution flow resumes.

The watchdog timer is a parallel process and does not interfere with the execution flow. Just the invocation of the watchdog timer is something that is added to the regular flow. However, in at least two instances with a 3.11.4 factory, it was observed that the execution of `cvmfs_helper_funcs.sh` and `cvmfs_setup.sh` resulted in an unusually long delay in completing the execution of these scripts. 

As part of troubleshooting the long delay problem, it was found that removing the `run_with_timeout` invocations as well as disabling the invocations by setting the timeout to 0, there were no delays seen. In contrast, when the timeout function was present, delays were observed.

With the watchdog timer's timeout set to 120 seconds, an interesting behavior was observed -- Upon tracking the timestamps before and after the timer function invocations, when the watchdog timer function is invoked inside `detect_local_cvmfs` function, the following was seen:
```
Mon Jul 20 23:46:15 UTC 2026
Mon Jul 20 23:48:15 UTC 2026
```
which shows a timeout of 120 seconds.

But, with the watchdog timer function invoked in `perform_system_check` function, the following was seen:
```
Mon Jul 20 23:48:15 UTC 2026
Mon Jul 20 23:48:15 UTC 2026
```

Rather than completely removing the timeout, this PR changes the default from 120 seconds to timeout disabled. We found that it will be useful to leave the code in there in case it's needed since the operators can enable the timeout and further investigation can be conducted, should the problem of hanging sites is ever encountered.

### DIFF
--- a/creation/web_base/cvmfs_helper_funcs.sh
+++ b/creation/web_base/cvmfs_helper_funcs.sh
@@ -16,7 +16,7 @@
 #LOGFILE="cvmfs_all.log"
 #exec &> $LOGFILE
 
-# Using run_with_timeout from glidein_paths.source sourced in glidein_startup.sh
+# Can use run_with_timeout from glidein_paths.source sourced in glidein_startup.sh to enable watchdog timer for sites where glidein enters a hanging state during pre-requisite system checks for on-demand CVMFS setup.
 
 variables_reset() {
 	# DESCRIPTION: This function lists and initializes the common variables
@@ -106,7 +106,9 @@ detect_local_cvmfs() {
 	CVMFS_ROOT="/cvmfs"
 	repo_name=oasis.opensciencegrid.org
 	# Second check...
-	if [[ -f "$CVMFS_ROOT/$repo_name"/.cvmfsdirtab || "$(run_with_timeout -s KILL 120 ls -A "$CVMFS_ROOT/$repo_name")" ]] &>/dev/null
+    # uncomment the following line instead to enable the use of watchdog timer
+    # if [[ -f "$CVMFS_ROOT/$repo_name"/.cvmfsdirtab || "$(run_with_timeout -s KILL 0 ls -A "$CVMFS_ROOT/$repo_name")" ]] &>/dev/null
+	if [[ -f "$CVMFS_ROOT/$repo_name"/.cvmfsdirtab || "$(ls -A "$CVMFS_ROOT/$repo_name")" ]] &>/dev/null
 	then
 		loginfo "Validating CVMFS with ${repo_name}..."
 		true
@@ -177,7 +179,9 @@ perform_system_check() {
     unshare -U true &>/dev/null
     GWMS_IS_UNPRIV_USERNS_ENABLED=$?
 
-    run_with_timeout -s KILL 120 check_fuse_installed
+    # uncomment the following line to enable the use of watchdog timer
+    # run_with_timeout -s KILL 0 check_fuse_installed
+    check_fuse_installed
     GWMS_IS_FUSE_INSTALLED=$?
 
     if [[ $GWMS_OS_VERSION_MAJOR -ge 9 ]]; then
--- a/creation/web_base/glidein_paths.source
+++ b/creation/web_base/glidein_paths.source
@@ -27,12 +27,12 @@ robust_realpath() {
     fi
 }
 
-run_with_timeout () {
+run_with_timeout() {
     # run_with_timeout [-s SIGNAL] [TIMEOUT] command, options and arguments (order is important)
     # if $1=-s, $2 signal to use to kill after timeout (otherwise the next will be $1)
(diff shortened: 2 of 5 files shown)


creation/web_base/glidein_paths.source
```diff
@@ -27,15 +27,20 @@ robust_realpath() {
...
+    local timeout
     local sig=TERM
     if [[ $1 = "-s" ]]; then sig=$2; shift 2; fi
-    if [[ $1 =~ ^[0-9]+$ ]]; then timeout=$1; shift; fi
+    if [[ ! $1 =~ ^[0-9]+$ ]]; then
+        timeout=0
+    else
+        timeout=$1
+        shift
+    fi
```
mambelli: `local timeout=0` on line 35 is sufficient

creation/web_base/cvmfs_helper_funcs.sh
```diff
@@ -106,7 +106,7 @@ detect_local_cvmfs() {
 	CVMFS_ROOT="/cvmfs"
 	repo_name=oasis.opensciencegrid.org
 	# Second check...
-	if [[ -f "$CVMFS_ROOT/$repo_name"/.cvmfsdirtab || "$(run_with_timeout -s KILL 120 ls -A "$CVMFS_ROOT/$repo_name")" ]] &>/dev/null
+	if [[ -f "$CVMFS_ROOT/$repo_name"/.cvmfsdirtab || "$(run_with_timeout -s KILL 0 ls -A "$CVMFS_ROOT/$repo_name")" ]] &>/dev/null
```
mambelli: just remove the timeout. Or leave also the commented line w/ timeout if you prefer

creation/web_base/cvmfs_helper_funcs.sh
```diff
@@ -177,7 +177,7 @@ perform_system_check() {
     unshare -U true &>/dev/null
     GWMS_IS_UNPRIV_USERNS_ENABLED=$?
 
-    run_with_timeout -s KILL 120 check_fuse_installed
+    run_with_timeout -s KILL 0 check_fuse_installed
```
mambelli: just remove the timeout. Or leave also the commented line w/ timeout if you prefer

mambelli (changes requested):
Should also set GWMS_CUSTOM_SCRIPTS_TIMEOUT to 0 in glidein_startup.sh and add and document a variable GLIDEIN_CUSTOM_SCRIPTS_TIMEOUT to allow to control the timeout value

creation/web_base/cvmfs_helper_funcs.sh
namrathaurs: Removed the `run_with_timeout()` invocation and left a commented version of the line for reference/future use.

creation/web_base/cvmfs_helper_funcs.sh
namrathaurs: Removed the `run_with_timeout()` invocation and left a commented version of the line for reference/future use.

creation/web_base/glidein_paths.source
namrathaurs: updated

namrathaurs:
> Should also set GWMS_CUSTOM_SCRIPTS_TIMEOUT to 0 in glidein_startup.sh and add and document a variable GLIDEIN_CUSTOM_SCRIPTS_TIMEOUT to allow to control the timeout value

Looks like the documentation for `GLIDEIN_CUSTOM_SCRIPTS_TIMEOUT` variable currently exists in the [production documentation](https://glideinwms.fnal.gov/doc.prd/factory/custom_vars.html#factory_vars). It has been listed under the section `Factory or Frontend variables (config - attr tags)`. Below is a screenshot of the same:


Screenshot 2026-08-19 at 09 35 18<br/>


The existing description has been updated to include a mention of the default value for `GWMS_CUSTOM_SCRIPTS_TIMEOUT` and that the config-defined `GLIDEIN_CUSTOM_SCRIPTS_TIMEOUT` variable allows control over the `GWMS_CUSTOM_SCRIPTS_TIMEOUT` variable that is hardcoded in the script.

mambelli (approved):

PR #696, opened 2026-07-29
V311/frontend credential debug

### LINKED ISSUE #695: Frontend warning is misleading when credentials are invalid (expired proxy/token)
## Summary
The frontend warning

"No security credentials match for factory pool ..., not advertising request; ... trust_domain and type ..."

is misleading when credentials are present but invalid (for example, an expired proxy).

## Observed behavior
In a real 3.11 deployment, the frontend emitted the mismatch warning even though trust_domain/type/auth_method were aligned.

Root cause was an expired proxy credential:

```text
voms-proxy-info -file /var/lib/gwms-frontend/.globus/x509_cmspilot02_cms_pilot.proxy
...
timeleft  : 0:00:00
```

Because invalid credentials are excluded during auth-method matching, the resulting warning looks like a config mismatch even when the actual issue is credential validity.

## Why this is confusing
Operators are directed toward trust_domain/type typos, while the immediate actionable problem may be token/proxy expiration.

## Proposed improvement
Update warning text to explicitly include invalid credential possibility, e.g.:

- mention "No usable security credentials" instead of "No security credentials match"
- mention that invalid/expired credentials can trigger this path

Also include a compact credential debug summary in logs:
- credential id/path/type/trust_domain/purpose/valid/invalid_reason
- required factory trust_domain/auth_method for the request

## Suggested impact
- Faster triage in production
- Fewer false config investigations when credentials are simply expired/not yet valid

mmascher:


frontend/glideinFrontendInterface.py
```diff
@@ -1438,10 +1466,17 @@ def do_advertise_one(
...
                 filename_arr = []  # don't try to advertise
+                req_name = params_obj.request_name
+                factory_trust, factory_auth = self.factory_constraint.get(req_name, ("<unknown>", "<unknown>"))
+                logSupport.log.warning(
+                    "No usable security credentials for factory pool %s, not advertising request;"
+                    " this may be a trust_domain/type mismatch, or invalid credentials (e.g. expired token/proxy)."
+                    " Check frontend credential trust_domain/type and credential validity against factory trust_domain/auth_method"
+                    % factory_pool
+                )
                 logSupport.log.warning(
```
mambelli, line 1477: This should be debug level.

frontend/glideinFrontendInterface.py
```diff
@@ -1399,6 +1399,34 @@ def do_advertise(self, file_id_cache=None, adname=None, create_files_only=False,
...
+            return "none"
+
+        details = []
+        for req_cred in self.request_credentials:
+            cred = req_cred.credential
+            try:
+                cred_id = cred.id
+            except Exception:  # credential may be uninitialized or invalid
+                cred_id = "<unavailable>"
+            invalid_reason = cred.invalid_reason()
```
mambelli, line 1414: I think copilots suggestions are correct. This should be protected in try except and valid derived from invalid reason or at least calculated here in the try-except

mambelli (changes requested):

frontend/glideinFrontendInterface.py
mmascher: Addressed in https://github.com/glideinWMS/glideinwms/commit/e4ce498132458ea995e0767b49d29c234767002f: moved the detailed credential debug dump from warning to debug level in the NoCredentialException path.

frontend/glideinFrontendInterface.py
mmascher: fixed with: https://github.com/glideinWMS/glideinwms/pull/696/commits/27459ab3116e764fcd96c44bf9bb4ee9977d81ce

mmascher:
Requested changes done.

mambelli:
Would be better to move the debug string generation to the Credential and RequestCredential classes.
Then it is not sure if the message addresses correctly the credentials in presence of snapshots (per-entry tokens).
Last the message may be better suited in `createAdvertiseWorkFile` to avoid duplicate or contradicting messages.
Check PR #706 for an alternative proposal.

PR #698, opened 2026-07-30
frontend: format child stderr/stdout logs for readability

### LINKED ISSUE #697: Frontend logs child tracebacks as raw bytes repr (hard to read)
## Summary
Frontend group child-process errors are currently logged as raw Python bytes repr (e.g. `b'Traceback ...\n'`), which is hard to read and slows debugging.

## Current behavior
When a group subprocess writes to stderr/stdout, parent frontend logging emits escaped byte strings in a single line, for example:

```text
WARNING: [group-name]: b'Traceback (most recent call last):\n ...'
```

This makes stack traces noisy and difficult to parse quickly.

## Impact
- Harder incident triage in production
- Operators must manually mentally decode escaped newlines
- Important error context is easy to miss

## Expected behavior
- Decode child stream bytes to text (UTF-8 with replacement fallback)
- Log as readable multiline blocks preserving traceback formatting
- Optionally cap max lines/chars and mark truncation to avoid log flooding

Example:

```text
WARNING: [group-name] STDERR:
    ...
```

## Suggested location
- `frontend/glideinFrontend.py`
- `poll_group_process()` stream logging path

## Reproduction context
Observed while troubleshooting frontend group failures (permission and credential-loading errors), where the real traceback was present but wrapped in unreadable bytes repr form.

mmascher:


mmascher:
This is now the formatted message from https://github.com/glideinWMS/glideinwms/issues/697#issuecomment-5129158058:

```
[2026-07-30 11:39:39,825] WARNING: [overflow] STDERR:
    File "/usr/lib/python3.9/site-packages/glideinwms/lib/credentials/credentials.py", line 723, in create_credential
      return credential_class(**kwargs)
    File "/usr/lib/python3.9/site-packages/glideinwms/lib/credentials/credentials.py", line 202, in __init__
      self.load(string, path)
    File "/usr/lib/python3.9/site-packages/glideinwms/lib/credentials/credentials.py", line 356, in load
      self.load_from_file(path)
    File "/usr/lib/python3.9/site-packages/glideinwms/lib/credentials/credentials.py", line 335, in load_from_file
      with open(path, "rb") as cred_file:
  PermissionError: [Errno 13] Permission denied: '/var/lib/gwms-frontend/.globus/x509_cmspilot02_cms_pilot.proxy'
  
  The above exception was the direct cause of the following exception:
  
    File "/usr/lib/python3.9/site-packages/glideinwms/frontend/glideinFrontendElement.py", line 2695, in <module>
      rcm = gfe.main()
    File "/usr/lib/python3.9/site-packages/glideinwms/frontend/glideinFrontendElement.py", line 334, in main
      self.configure()
    File "/usr/lib/python3.9/site-packages/glideinwms/frontend/glideinFrontendElement.py", line 275, in configure
      group_dir, glideinFrontendPlugins.createRequestBundle(self.elementDescript.merged_data)
    File "/usr/lib/python3.9/site-packages/glideinwms/frontend/glideinFrontendPlugins.py", line 806, in createRequestBundle
      security_bundle.load_from_element(elementDescript)
    File "/usr/lib/python3.9/site-packages/glideinwms/lib/credentials/utils.py", line 78, in load_from_element
      credential = create_credential(
    File "/usr/lib/python3.9/site-packages/glideinwms/lib/credentials/credentials.py", line 727, in create_credential
      raise CredentialError(f'Unexpected error loading credential: string="{string}", path="{path}"') from err
  glideinwms.lib.credentials.credentials.CredentialError: Unexpected error loading credential: string="None", path="/var/lib/gwms-frontend/.globus/x509_cmspilot02_cms_pilot.proxy"
[2026-07-30 11:39:39,836] WARNING: [local_users] STDERR:
```

frontend/glideinFrontend.py
```diff
@@ -128,6 +128,37 @@ def spawn_group(work_dir, group_name, action):
...
+        text = payload.decode("utf-8", errors="replace")
+    else:
+        text = str(payload)
+
+    text = text.strip()
+    if not text:
+        return None
+
+    if len(text) > max_chars:
+        text = text[:max_chars] + "\n... output truncated ..."
```
namrathaurs, line 150: Any particular reason why the truncation logic is introduced rather than writing/using the full error stack trace?

frontend/glideinFrontend.py
```diff
@@ -128,6 +128,37 @@ def spawn_group(work_dir, group_name, action):
...
+    """Format child stream output for readable logs.
+
+    Decodes bytes, keeps multiline formatting, and truncates very long output.
+    """
+    if not payload:
+        return None
+
+    if isinstance(payload, bytes):
+        text = payload.decode("utf-8", errors="replace")
+    else:
```
namrathaurs, line 142: 1. What cases are covered under the `else` block? I'm assuming every other data type, except `str`. 
If my assumption is correct, wondering if that is even a possibility when dealing with output/error streams.

2. Do we also have to explicitly handle the case of `payload` being of `str` type here? Is that even possible since we are dealing with output/error stream here?

namrathaurs (commented):
Some clarifications needed.

frontend/glideinFrontend.py
mmascher: About 1, you are probably right that we are only dealing with streams and the else is not relevant. But considering this was the behavior before the change (=> `f"[{group_name}]: {tempOut}"`), I'd like to keep the else just to be safe. It also covers 2, i.e., the case where payload is str.

namrathaurs: Thanks for the clarification; marking this as resolved.

frontend/glideinFrontend.py
namrathaurs: discussed this with @mmascher and it was agreed to retain the full length of the error stack without truncation. waiting for the update to reflect before the changes can be approved.

namrathaurs: this comment has been addressed and the truncation logic has been removed.

namrathaurs (approved):
Comments have been addressed; changes can be merged.

PR #701, opened 2026-08-11
Adding known issues to CHANGELOG

### LINKED ISSUE #679: Error in scripts verifying x509

### LINKED ISSUE #681: Factory's `schedd_status.xml` doesn't report existing pilots after entry `auth_method` changes
**Describe the bug**
After changing the credentials for a factory entry (e.g. changing the `auth_method` from `grid_proxy` to `scitoken`), `schedd_status.xml` stops reporting pilots using the previous credentials.

I'm not sure if this an actual bug or if it's by design, but this has caught my attention when changing the `auth_method` of entries in the CERN Factory.

**To Reproduce**

1. Setup a factory entry with `auth_method="grid_proxy"`
2. Create pressure to trigger the submission of pilots to the CE
3. Verify the number of running pilots reported in the `schedd_status.xml` for the entry
4. Change the entry to use `auth_method="scitoken"`
5. Reconfigure the factory
6. Verify the number of running pilots reported in the `schedd_status.xml` for the entry. Only pilots submitted with the new credential are shown
7. Query the Factory schedd with `condor_q -all -global -constraint 'GlideinEntryName =?= "<entry-name>" && GlideinCredentialIdentifier =!= UNDEFINED' -af JobStatus GridJobStatus GlideinCredentialIdentifier` to verify that pilots with the previous credentials are still running, but are not reported in `schedd_status.xml`

**Expected behavior**
Running pilots with old credentials should still be reported in `schedd_status.xml`.

**Screenshots**
At CERN we extract the data from `schedd_status.xml` and plot it in Grafana. This plot shows the described behavior over time for one entry. The `auth_method` of the entry was changed at around 10:12.

Image

-   GlideinWMS version: 3.11.3-2
-   Python version: 3.9.25
-   OS version: AlmaLinux 9.8
-   HTCondor version: 24.0.20
-   Priority: low
-   Stakeholders: CMS, FactoryOps
-   Components: factory

**Additional context**

Here's a full example with a real entry at CERN, the same from the screenshot above. Here we can see that `schedd_status.xml` reports only the pilots with the new credential ID `W5JFFRTO`.

```shell
$ cat /var/lib/gwms-factory/work-dir/monitor/entry_CMSHTPC_T2_CH_CERN_ce513-beer/schedd_status.xml
<?xml version="1.0" encoding="ISO-8859-1"?>

<glideFactoryEntryQStats>
   <updated>
      <timezone name="Local" ISO8601="2026-06-24T14:51:39+02:00" RFC2822="Wed, 24 Jun 2026 14:51:39 +0200" human="Wed Jun 24 14:51:39 2026"/>
      <timezone name="UTC" ISO8601="2026-06-24T12:51:39Z" RFC2822="Wed, 24 Jun 2026 12:51:39 +0000" unixtime="1782305499"/>
   </updated>
   <downtime status="False">
   </downtime>
   <frontends>
      <frontend name="CMS_T0-Frontend_cmspilot">
         <ClientMonitor CoresIdle="0" CoresRunning="16" CoresTotal="16" GlideIdle="0" GlideRunning="2" GlideTotal="4" InfoAge="88" JobsIdle="324546" JobsRunHere="2" JobsRunning="32284"/>
         <Downtime status="False"/>
         <Requested AdjustedIdle="20" Idle="20" IdleCores="160" MaxCores="178464" MaxGlideins="22308">
            <Parameters>
            </Parameters>
         </Requested>
         <Status Held="0" Idle="10" IdleOther="0" Pending="10" Running="2" RunningCores="16" StageIn="0" StageOut="0" Wait="0"/>
         <StatusEntries/>
      </frontend>
   </frontends>
   <total>
      <ClientMonitor CoresIdle="0" CoresRunning="16" CoresTotal="16" GlideIdle="0" GlideRunning="2" GlideTotal="4" InfoAge="88" JobsIdle="324546" JobsRunHere="2" JobsRunning="32284"/>
      <Requested AdjustedIdle="20" Idle="20" IdleCores="160" MaxCores="178464" MaxGlideins="22308"/>
      <Status Held="0" Idle="10" IdleOther="0" Pending="10" Running="2" RunningCores="16" StageIn="0" StageOut="0" Wait="0"/>
   </total>
</glideFactoryEntryQStats>

$ condor_q -all -global -constraint 'GlideinEntryName =?= "CMSHTPC_T2_CH_CERN_ce513-beer"' -af JobStatus GridJobStatus GlideinCredentialIdentifier
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING FFOGCMY5
2 RUNNING W5JFFRTO
2 RUNNING W5JFFRTO
1 IDLE W5JFFRTO
1 IDLE W5JFFRTO
1 IDLE W5JFFRTO
1 IDLE W5JFFRTO
1 IDLE W5JFFRTO
1 IDLE W5JFFRTO
1 IDLE W5JFFRTO
1 IDLE W5JFFRTO
1 IDLE W5JFFRTO
1 IDLE W5JFFRTO
```

**An initial investigation attempt, might be inaccurate**

I looked a bit into it and I see that [the pilots are filtered by the credential ID](https://github.com/glideinWMS/glideinwms/blob/87fcec1c79b4a2c50c4b6c481f840d1fafd6b32d/factory/glideFactoryLib.py#L337), which IIUC is derived from the credential file name.

namrathaurs:
This PR adds two issues that were recently identified as known issues upon thorough investigations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ Changes since the last release
 
 ### Known Issues
 
+-   Factory monitoring does not report existing pilots after entry `auth_method` changes from `grid_proxy` to `scitoken` and `CONTINUE_IF_NO_PROXY` knob is unset (Issue #681)
+-   Running 3.10.x Frontend continues to request Glideins even when unable to create IDTOKENS, resulting in Glideins failing at startup (Issue #679, PR #680)
+
 ## v3.10.18 \[2026-04-17\]
 
 Added custom scripts timeout, fixed job monitoring and few other bugs.


CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ Changes since the last release
 
 ### Known Issues
 
+-   Factory monitoring does not report existing pilots after entry `auth_method` changes from `grid_proxy` to `scitoken and `CONTINUE_IF_NO_PROXY` knob is unset (Issue #681)
```
mambelli: Missing quote after scitoken.

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ Changes since the last release
...
+-   Absence of Glidein IDTOKEN in 3.10.x Frontend and cached Glidein IDTOKEN in 3.11.x Factory results in glidein startup failure (Issue #679, PR #680)
```
mambelli: something like "Running 3.10.x Frontend continues to request Glideins also when unable to create IDTOKENS, resulting in Glideins failing at startup" 
I think this happens with both 3.10 and 3.11 Factories

mambelli (changes requested):

CHANGELOG.md
namrathaurs: fixed!

CHANGELOG.md
namrathaurs: That reads better, thanks for the suggestion. Updated to the suggested version.

My investigation did not specifically test for 3.10 Frontend with 3.10 Factory, but I believe what you said is correct about this being problematic with both 3.10. and 3.11 Factories since the Frontend behavior is what causes this issue.

mambelli (approved):

PR #704, opened 2026-08-20
Retrieving GLIDEIN_SINGULARITY_BINDPATHs from glidein_config

### LINKED ISSUE #703: GLIDEIN_SINGULARITY_BINDPATH ignored in custom scripts
**Describe the bug**
GLIDEIN_SINGULARITY_BINDPATH is not used for custom scripts running in containers.

@rynge reported: With a config like this:
```
      <attr name="GLIDEIN_SINGULARITY_BINDPATH" glidein_publish="True" job_publish="True" parameter="False" type="string" value="/alphafold3,/mnt/alphafold,/mnt/alphafold/alphafold3_7a4a2f,/scitech/nas/scratch-90-days/alphafold3"/>
```
GLIDEIN_SINGULARITY_BINDPATH is not used used when running the scripts in the default container:
```
<file absfname="/opt/osg-flock/ospool-pilot/main/pilot/advertise-userenv" type="run:singularity" after_entry="True" const="True" executable="True" prefix="NOPREFIX" untar="False" wrapper="False">
```

**To Reproduce**
Have a custom script running in Apptainer. 

**Expected behavior**
Add the paths in GLIDEIN_SINGULARITY_BINDPATH

**Screenshots**

-   GlideinWMS version: 3.10.x, 3.11.x
-   Python version:
-   OS version:
-   HTCondor version:
-   Priority: high
-   Stakeholders: OSG

**Additional context**
The variable works for the jobs but not the custom script (which they use the same library functions), this only when ` job_publish="True"`. The problem is that the variable is retrieved only from the environment and not glidein_config.

mambelli:
Retrieving GLIDEIN_SINGULARITY_BINDPATH and GLIDEIN_SINGULARITY_BINDPATH_DEFAULT from glidein_config instead of the environment. 
This way both are always used in both jobs and custom scripts, not just when in the environment.

This fixes #703

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,6 @@ Changes since the last release
 
 ### New features / functionalities
 
--   item one of the list
--   item N
-
 ### Changed defaults / behaviours
 
 ### Deprecated / removed options and commands
@@ -21,6 +18,8 @@ Changes since the last release
 ### Bug Fixes
 
 -   Fix incorrect Arch requirement for ARM pilots (Issue #661, PR #662)
+-   Sanitize cvmfs repos list to protect against Arbitrary Code Execution attack (PR #678)
+-   Retrieve GLIDEIN_SINGULARITY_BINDPATH from glidein_config instead of the environment so it is not ignored (Issue #703, PR #704)
 
 ### Testing / Development
 
--- a/creation/web_base/singularity_lib.sh
+++ b/creation/web_base/singularity_lib.sh
@@ -1202,14 +1202,17 @@ singularity_get_binds() {
     local retv=  # default controlled from outside ($2)
     local checks=$1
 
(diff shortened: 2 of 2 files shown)


mambelli:
To path either 3.10.18 or 3.11.4, you can replace singularity_lib in the web-base directory (/var/lib/gwms-factory/web-base) with the version in this PR https://raw.githubusercontent.com/mambelli/glideinwms/refs/heads/v310/fix_bindpathretrieval/creation/web_base/singularity_lib.sh
and issue a factory upgrade command

shreyb (approved):
This looks good to me.  If I understand correctly, rather than reading `$GLIDEIN_SINGULARITY_BINDPATH{_DEFAULT}` from the environment, we're pulling those values from the GWMS config?

mambelli:
> This looks good to me. If I understand correctly, rather than reading `$GLIDEIN_SINGULARITY_BINDPATH{_DEFAULT}` from the environment, we're pulling those values from the GWMS config?

Correct. glidein_config is the file used to exchange values between all the custom scripts (the scripts invoked by the glidein)

PR #705, opened 2026-08-26
Improve request credentials match troubleshooting

### LINKED ISSUE #695: Frontend warning is misleading when credentials are invalid (expired proxy/token)

mambelli:
This PR improves the misleading warning in `do_advertise_one()` mentioning the possibility of expired credentials as culprit for the failure to find request credentials for a given request.

It also fixes the JWT credential that could throw exceptions due to missing claims when checking its validity.

Furthermore it improves the docstrings in `glideinFrontendLibrary.py`

This fixes #695 and supersedes #696

It is currently in draft, while deciding if the debug messages should be moved to `createAdvertiseWorkFile()`, where the exception is risen. Also to decide which debug message to use, the one with or without snapshot.


mambelli:
Wrong base branch, superseded by #706

PR #707, opened 2026-08-26
Switched cat_consts to add lines in batches

rynge:
We are still seeing timeouts glidein startup timeouts on some sites with slower filesystems. The extreme case of this is always failing in cat_consts.sh with something like:

```
Signature OK for main:cat_consts.q8pfwd.sh.
=== Validation error in /mnt/nfs/scratch/osg01/glide_qbjN9X/main/cat_consts.sh ===
Command /mnt/nfs/scratch/osg01/glide_qbjN9X/main/cat_consts.sh killed by timeout (600 s)
```

Part of the problem is the large number of constants being passed in. This mostly comes from the many different HTCondor tarballs we have to support, and it seems like there is nothing we can do about that.

This PR provides a performance improvement for cat_consts.sh, by batching the update. Details:

1. creation/web_base/add_config_line.source - added gconfig_add_multi(). It does the equivalent of gconfig_add() for a whole batch of "key value" lines (read from stdin), but copies/greps/renames glidein_config once for the whole batch instead of once per key. It preserves the existing semantics: last value wins per key (both within the batch and vs. pre-existing entries), and still logs everything to glidein_config.history. gconfig_add(), gconfig_add_safe(), etc. are untouched.
2. creation/web_base/cat_consts.sh - instead of calling gconfig_add once per line while looping over the entry's consts_file, it now accumulates all the lines and does a single gconfig_add_multi call at the end. nr_lines counting is preserved (the loop stays out of any subshell/pipe so it can still increment in place).

Why this addresses your timeout: the old loop did a full grep + cp + grep + mv of the (growing) glidein_config file per constant. On a site like this one where the whole glidein sandbox lives on a slow shared filesystem (including OSG_WN_TMP), each of those file ops is a network round trip, and with dozens/hundreds of constants that adds up fast - easily enough to blow past the 600s script timeout that killed cat_consts.sh in your log, even though the same entry runs fine on local-disk sites.

### DIFF
--- a/creation/web_base/add_config_line.source
+++ b/creation/web_base/add_config_line.source
@@ -5,7 +5,11 @@
 
 # IMPORTANT
 # These functions use the variable $glidein_config
-# gconfig_add() (and gconfig_add_unsafe and gconfig_add_safe) all require $glidein_config, otherwise they exit
+# gconfig_add() (and gconfig_add_unsafe, gconfig_add_safe and gconfig_add_multi) all require $glidein_config, otherwise they exit
+# gconfig_add_multi() adds many "key value" lines (read from stdin, one per line) in a single
+#  read-modify-write pass, instead of one full glidein_config rewrite per call like gconfig_add().
+#  Use it instead of looping over gconfig_add() when adding more than a few lines at once
+#  (e.g. merging in a constants file), especially on slow/networked filesystems.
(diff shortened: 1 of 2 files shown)


mambelli (approved):

mambelli:
@rynge Thanks for the great PR.
I made a couple of minor changes and merged it:
- modified slightly the temp file names to make more clear that they referred to the added lines.
- added a TODO note to cat_consts.sh to suggest future improvements

PR #711, opened 2026-09-02
Improved generators context validation and code documentation

mambelli:
Improved the generators context validation and code documentation. 
- Added the `force_value` function generator to help with context validation
- Modified all credential plugins so that the context is validated also at reconfig time
- Improved also the operators documentation


mmascher (dismissed):
Looks good to me.

mmascher (approved):
LGTM

PR #716, opened 2026-09-11
Standardize reading/writing of tokens to remove/add newline characters

### LINKED ISSUE #639: Make sure token are written w/ newline and stripped at reading

All writing and reading of JWTs (IDTOKENS, SciTokens, ...) should be consistent:
- when writing a newline should be added at the end
- when reading the string should be stripped to remove newlines and spaces before and after

Consider if this should be done also w/ other credentials.

Checking w/ people, this is what they expect:

```
I think users of tokens should work with either.  htgettoken puts a newline in.
Dave
```

```
Hi Marco,
I personally don't have enough knowledge to make a strong opinion on this matter, but I did pass things along to the CHTC team and got the following opinions (from a couple of people including Jaime our HTCSS Security person):
Add a newline when writing a new token to a file (so it's easier to cat  later)
Accept tokens whether they have a newline at the end or not. Actually, just ignore all whitespace on either side of the token.
This is called for in the [WLCG token discovery spec](https://zenodo.org/records/3937438)
Not sure if it matters, but Jaime did also point out that HTCondor binaries will properly handle IDTOKEN files that do not contain a newline.
Cheers,
Cole Bollig
```

NA

-   Priority: medium
-   Stakeholders: Fermilab, OSG
-   Components: frontend, factory, glidein

**Additional context**

Check the refactored credential code and maybe the glidein shell code (if using the tokens, probably it is just copying them around and passing them to other programs)

shreyb:
Closes #639.

This PR adds a newline character (`\n`) anywhere a token is written to a file, that didn't already do so, and strips whitespace (including `\n`) when reading tokens.

### DIFF
--- a/creation/web_base/update_proxy.py
+++ b/creation/web_base/update_proxy.py
@@ -122,6 +122,7 @@ def main():
                 with open(idtokens_file) as idtf:
                     for line in idtf.readlines():
                         idtoken_data += line
+            idtoken_data.strip()
             compressed_credential = compress_credential(credential_data)
             update_credential(fname_compressed, f"{idtoken_data}####glidein_credentials={compressed_credential}")
             # in branch_v3_2 after migration_3_1 WAS: update_credential(fname_compressed, compressed_credential)
--- a/factory/glideFactoryEntry.py
+++ b/factory/glideFactoryEntry.py
@@ -1162,6 +1162,7 @@ def unit_work_v3(
     condortoken_file = os.path.join(submit_credentials.cred_dir, condortokenbase)
     condortoken_data = decrypted_params.get(condortoken)
     if condortoken_data:
+        condortoken_data = f"{condortoken_data.strip()}\n"
         fd, tmpnm = tempfile.mkstemp(dir=submit_credentials.cred_dir)
         try:
             entry.log.info(f"frontend_token supplied, writing to {condortoken_file}")
--- a/frontend/glideinFrontendElement.py
+++ b/frontend/glideinFrontendElement.py
@@ -1236,7 +1236,7 @@ class glideinFrontendElement:
                     # The token file is read as text file below. Writing fixed to be consistent
                     with tempfile.NamedTemporaryFile(mode="w", delete=False, dir=tkn_dir) as fd:
                         os.chmod(fd.name, 0o600)
-                        fd.write(tkn_str)
+                        fd.write(f"{tkn_str.strip()}\n")
                         os.replace(fd.name, tkn_file)
                     logSupport.log.debug("created token %s" % tkn_file)
                 elif os.path.exists(tkn_file):
--- a/logserver/getjwt.py
+++ b/logserver/getjwt.py
@@ -87,7 +87,7 @@ else:
     try:
         # Write the token to a text file
         with open(token_filepath, "w") as tkfile:
-            tkfile.write(token)
+            tkfile.write(f"{token}\n")
         log(f"Token for {args.log_url} ({urllib.parse.quote(args.log_url, '')}) written to {token_filepath}")
     except OSError:
         log(f"ERROR: Unable to create JWT file: {token_filepath}")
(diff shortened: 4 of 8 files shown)


creation/web_base/update_proxy.py
```diff
@@ -122,6 +122,7 @@ def main():
                 with open(idtokens_file) as idtf:
                     for line in idtf.readlines():
                         idtoken_data += line
+            idtoken_data.strip()
```
namrathaurs: More of a curious question: does the indentation level matter here? I noticed that the `factory/glideFactoryCredentials.py` file (probably right after this one) in this PR has the same line of code but indented at a much inner level (at the same level as `for`)?

frontend/glideinFrontendElement.py
```diff
@@ -1236,7 +1236,7 @@ def refresh_entry_token(self, glidein_el):
                     # The token file is read as text file below. Writing fixed to be consistent
                     with tempfile.NamedTemporaryFile(mode="w", delete=False, dir=tkn_dir) as fd:
                         os.chmod(fd.name, 0o600)
-                        fd.write(tkn_str)
+                        fd.write(f"{tkn_str.strip()}\n")
```
namrathaurs: Is the idea here to remove all the leading and trailing whitespace characters AND just add a `\n` at the end of the string?

logserver/getjwt.py
```diff
@@ -87,7 +87,7 @@ def log(msg: str):
     try:
         # Write the token to a text file
         with open(token_filepath, "w") as tkfile:
-            tkfile.write(token)
+            tkfile.write(f"{token}\n")
```
namrathaurs: Any particular reason why the strip function was not used here (as in `f"token.strip()\n"`)?

If considered to add, does it provide any value?

namrathaurs (commented):
Changes look fine. Some minor comments requesting clarifications.

frontend/glideinFrontendElement.py
shreyb: Yes - exactly that.

creation/web_base/update_proxy.py
mambelli: No big difference, but I'd put it at the for level, or at least with. No need to strip the default empty string

factory/glideFactoryEntry.py
```diff
@@ -1162,6 +1162,7 @@ def unit_work_v3(
     condortoken_file = os.path.join(submit_credentials.cred_dir, condortokenbase)
     condortoken_data = decrypted_params.get(condortoken)
     if condortoken_data:
+        condortoken_data = f"{condortoken_data.strip()}\n"
```
mambelli: No strip here. The data should be stripped before, at reading form file or message. In memory should be stripped, otherwise "if condortoken_data" would not work, accepting only spaces as a token

frontend/glideinFrontendElement.py
mambelli: The string is a variable in memory and should be already stripped. No need for strip here, check when `tkn_str` is read from file/message

logserver/getjwt.py
mambelli: ok not to have it (see above)

logserver/jwt.php
```diff
@@ -61,8 +61,8 @@
         $jwt = JWT::encode($payload_array, $key, 'HS256');
         print_r($jwt);
         echo "Saving token to $argv[2]\n";
-        file_put_contents($argv[2], $jwt);
-        $decoded = JWT::decode($jwt, new Key($key, 'HS256'));
+        file_put_contents($argv[2], $jwt . "\n");
+        $decoded = JWT::decode(trim($jwt), new Key($key, 'HS256'));
```
mambelli: No need for trim here, the variable should be already stripped

mambelli (changes requested):

